PR #8, opened 2023-07-21
refactor(bootstrap): generalize slide asset pre-caching

jpnurmi:
Don't assume that slide assets are in `flutter_assets/packages/ubuntu_bootstrap/assets/slides` because this won't be the case if `ubuntu_bootstrap` itself was made an app for development and testing purposes. In that case, the assets would be in `flutter_assets/assets/slides`. Furthermore, we should also pre-cache slide assets that have been overridden by flavors and OEMs, which is better achieved by iterating the asset manifest than looking into specific directories on the file system.


d-loose (approved):
LGTM :+1:

PR #13, opened 2023-07-24
fix(timezone): fix XDG default

jpnurmi:
While 'geoip' is the desired default for Subiquity, it's not a valid value for XDG so default to 'Etc/UTC' instead.

```
The following DBusInvalidArgsException was thrown running a test:
org.freedesktop.DBus.Error.InvalidArgs: Invalid or not installed time zone 'geoip'
```


d-loose (approved):
LGTM :+1:

PR #14, opened 2023-07-24
feat(init): add network

jpnurmi:
The network page does not depend on Subiquity and thus, can be supported in the init-phase.


d-loose (approved):
LGTM :+1:

PR #19, opened 2023-07-25
test: ubuntu_init integration test

jpnurmi:
Runs with the keyfile backend for GSettings and a fake D-Bus server to make it safe to run the test in any environment.


d-loose (approved):
LGTM :+1:

PR #20, opened 2023-07-25
refactor: register GSettings factory

jpnurmi:
Allows wiring up custom D-Bus clients and keyfile backends for GSettings for testing purposes.

GSettings is somewhat complicated because it writes through DConf's D-Bus API but reads directly from the file system. Even though `gsettings.dart` allows setting environment variables for customization, this is not usable in tests because the immutable `Platform.environment` is already accessed and cached before the test entry point gets called so it's too late to set any environment variables at the beginning of tests.

See also:
- #19


d-loose (approved):
LGTM :+1:

PR #30, opened 2023-07-27
fix(init): model initialization on startup

jpnurmi:
Calling asynchronous `InitModel.init()` from within `InitWizard.initState()` leaves the future dangling. Make use of `WizardRoute.onReplace` from the initial loading route to ensure that the initialization routine finishes before proceeding to the next route.


d-loose (approved):
LGTM :+1:

PR #33, opened 2023-07-28
feat(config): look up the config file unless explicitly specified

jpnurmi:
Unless `--config <path>` is passed, look up the config file path in the following order:

- `/etc/ubuntu-provision.{conf,yaml,yml}` (admin)
- `/usr/local/share/ubuntu-provision.{conf,yaml,yml}` (oem)
- `/usr/share/ubuntu-provision.{conf,yaml,yml}` (distro)

Notes:
- The standard [XDG config and data dirs](https://pub.dev/packages/xdg_directories) are supported.
- To limit the amount of FS checks, only a subset of the supported extensions are looked up.

Ref: canonical/ubuntu-desktop-installer#2194


d-loose (approved):
LGTM :+1:

PR #34, opened 2023-07-28
chore: optimize melos & ci

jpnurmi:
- Run coverage jobs in parallel. The downside is to have to list them in the matrix but on the upside, they only take a few minutes each.
- Run all tests without coverage collection to ensure no test gets skipped if they are forgotten from the manually maintained coverage matrix. Executing all tests without coverage collection doesn't take much longer than it takes to run the longest test with coverage collection.
- Ignore `subiquity_client` & `subiquity_test` units tests because they are already run at [subiquity_client.dart](https://github.com/canonical/subiquity_client.dart). This saves us from installing subiquity dependencies for running unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,36 @@ jobs:
     - run: melos pub get
     - run: melos analyze
 
+  coverage:
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


d-loose (approved):
LGTM :+1:

PR #35, opened 2023-07-28
test: import bootstrap integration test from UDI

jpnurmi:
We need to run integration tests for `ubuntu_bootstrap` to ensure that we find potential issues here and not in downstream repos such as UDI or flavors.

Ref: canonical/ubuntu-desktop-installer#2194


d-loose (approved):
LGTM :+1:

PR #37, opened 2023-07-28
fix(provision): broadcast journal stream

jpnurmi:
In Dart, generator streams are always single-subscription streams, even if it yields a broadcast stream, and a single-subscription stream does not allow closing and re-subscribing.

The journal stream for the log view must be a broadcast stream to be able to resubscribe in case of state changes and rebuilds. The issue was noticed in #35:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following StateError was thrown building _JournalView(dependencies: [_InheritedTheme,
_LocalizationsScope-[GlobalKey#05e92]]):
Bad state: Stream has already been listened to.

The relevant error-causing widget was:
  _JournalView
  _JournalView:file:///home/runner/work/ubuntu-desktop-provision/ubuntu-desktop-provision/packages/ubuntu_bootstrap/lib/pages/install/install_page.dart:104:26
```

### DIFF
--- a/packages/ubuntu_provision/lib/src/services/journal_service.dart
+++ b/packages/ubuntu_provision/lib/src/services/journal_service.dart
@@ -2,8 +2,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:async/async.dart';
-
 enum JournalOutput {
   /// The default and generates an output that is mostly identical to the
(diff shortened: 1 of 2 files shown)


d-loose (approved):
LGTM :+1:

PR #39, opened 2023-07-31
refactor(config): rename 'routes' to 'pages'

jpnurmi:
The term "page" is more user-friendly and matches with GIS, whereas "route" basically refers to an implementation detail.


d-loose (approved):
LGTM :+1:

PR #40, opened 2023-07-31
feat(config): introduce bootstrap & init scopes

jpnurmi:
Makes it possible to run different wizard stages with the same config:
```ini
[bootstrap]
pages = ...

[init]
pages = ...
```


d-loose (approved):
LGTM :+1:

PR #45, opened 2023-07-31
feat(config): allow apps to ship default config asset

jpnurmi:
If nothing else is configured on the system, allow specific `ubuntu_bootstrap` or `ubuntu_init` builds to configure which pages they should show by default.


d-loose (approved):
LGTM :+1:

PR #46, opened 2023-07-31
fix(config): add missing `assets/` subdir

jpnurmi:
We'll copy a config file into `assets/` during snapcraft build process. `ubuntu_bootstrap` and `ubuntu_init` will automatically pick it from there as they both have:
```yaml
flutter:
  assets:
    - assets/
```


d-loose (approved):
LGTM :+1:

PR #48, opened 2023-08-01
fix(wizard): don't disable the last next/done button

jpnurmi:
No individual pages should have to force-enable Next/Done. Any page could be configured as the last page. It must be possible to click Done to finish the wizard and close the window even if there's no next page route.


d-loose (approved):
LGTM :+1:

Issue #42, opened 2023-07-31
welcome: telemetry


jpnurmi:
![Set up your account](https://github.com/canonical/ubuntu-desktop-provision/assets/140617/75b5ce80-89a4-4c26-bc11-37396c1fc1a5)

jpnurmi:
![Telemetry](https://github.com/canonical/ubuntu-desktop-provision/assets/140617/9e2bce8f-6e9c-418c-b0cf-f1142bee1d39)

PR #65, opened 2023-08-07
feat(init): add libsysmetrics FFI bindings

jpnurmi:
For the telemetry page that determines whether ubuntu-report is used.
- #42


d-loose (approved):
LGTM :+1:

Issue #41, opened 2023-07-31
init: welcome


jpnurmi:
![Welcome](https://github.com/canonical/ubuntu-desktop-provision/assets/140617/54fd5df2-5992-4d63-898f-3d24cb09e068)

jpnurmi:
![Welcome (1)](https://github.com/canonical/ubuntu-desktop-provision/assets/140617/dc00de1e-9d7f-4ea8-ab04-9c399c1c4f86)

PR #66, opened 2023-08-07
refactor(provision): promote ProductService & release notes URL lookup

jpnurmi:
ProductService extracts the info from `/cdrom/.disk/info` (i.e. from the ISO in a live session) if it exists or falls back to `/etc/os-release` which works outside the live session ie. works great for `ubuntu_init` too, where it is needed for presenting the product version and changelog link on the welcome page:
- #41


d-loose (approved):
LGTM :+1:

PR #71, opened 2023-08-08
refactor(sysmetrics): move blocking FFI calls into isolates

jpnurmi:
It takes a significant amount of time to collect the data. Hide the FFI calls into isolates rather than forcing the telemetry view model to do that because it makes the sysmetrics calls non-mockable/verifiable.

[Screencast from 2023-08-08 14-09-47.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/140617/4d39ba62-5785-42e2-b7e8-9aa5056921f6)

Ref: #42


d-loose (approved):
LGTM :+1:

PR #73, opened 2023-08-08
feat(bootstrap): add minimal changes for dd targets

jpnurmi:
DD targets are handled the same way as other reformat targets. Additionally, advanced features are only offered if appropriate.

```
$ ubuntu_bootstrap --source-catalog=examples/sources/core.yaml
```

image


kenvandine (approved):
Seems straight forward, LGTM

PR #80, opened 2023-08-09
chore(test): add leading zeros to screenshot numbers

jpnurmi:
To fix sorting here and there, including GitHub where people often browse these screenshots:

image


d-loose (approved):
LGTM :+1:

PR #82, opened 2023-08-10
feat(init): allow passing app title & l10n delegates to runInitApp()

jpnurmi:
I missed that `ubuntu-welcome` has its own localized app title that it must pass in to be able to use `runInitApp()`.


d-loose (approved):
LGTM :+1:

PR #87, opened 2023-08-10
fix(store): remove can-launch check

jpnurmi:
`canLaunchUrl()` doesn't work from a confined snap even if `launchUrl()` does.


d-loose (approved):
LGTM :+1:

PR #91, opened 2023-08-11
feat(flavors): restore the ability to pass flavor data

jpnurmi:
Even if [ubuntu_flavor](https://pub.dev/packages/ubuntu_flavor) is able to detect most official Ubuntu flavors, there needs to be a way to pass a specific flavor for being able to:
- lock down a specific flavor regardless of the host platform (ubuntu-xxx-installer should always run as xxx)
- pass an imaginary custom flavor for demoing purposes (ubuntu-flavor-installer)


d-loose (approved):
LGTM :+1:

jpnurmi:
For example,
```dart
runInstallerApp(args, flavor: UbuntuFlavor.budgie);
```
image
(on a vanilla Ubuntu host with the default orange accent)

PR #92, opened 2023-08-11
fix(bootstrap): expose missing l10n arguments for runInstallerApp()

jpnurmi:
Same as #82 but for ubuntu_bootstrap, needed for flavors' translated installations slides.


d-loose (approved):
LGTM :+1:

PR #104, opened 2023-08-24
fix: Only show installation sources if there is more than one

jibel:
Sources are a radio button and it should not be displayed if there is just one source. In this case we go with the default.

closes: https://github.com/canonical/ubuntu-desktop-installer/issues/2305

### DIFF
--- a/packages/ubuntu_bootstrap/lib/pages/source/source_page.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/source/source_page.dart
@@ -23,6 +23,7 @@ class SourcePage extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final model = ref.watch(sourceModelProvider);
+    final showSources = model.sources.length > 1;
     final lang = UbuntuBootstrapLocalizations.of(context);
     return WizardPage(
       title: YaruWindowTitleBar(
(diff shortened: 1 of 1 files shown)


packages/ubuntu_bootstrap/lib/pages/source/source_page.dart
```diff
@@ -39,20 +39,26 @@ class SourcePage extends ConsumerWidget {
...
                     alignment: AlignmentDirectional.centerStart,
-                    child: YaruRadioButton<String>(
-                      title: Text(source.localizeTitle(context)),
-                      subtitle: Text(source.localizeSubtitle(context)),
-                      contentPadding: kWizardPadding,
-                      value: source.id,
-                      groupValue: model.sourceId,
-                      onChanged: model.setSourceId,
+                    child: Visibility(
+                      visible: model.sources.length > 1,
```
CarlosNihelton, line 43: I bet the compiler can optimize this out, but it'd rather store this once in the beginning of the `Widget build(...` method and reuse in here and in the other `Visibility` widget, thus avoiding the repetition.

```dart
final model = ref.watch(sourceModelProvider);
final showRadioButtons = model.sources.length > 1;
```

and then

```dart
                    child: Visibility(
                      visible: showRadioButtons,
...
```

jibel: I still need to compute the button because it is then used to retrieve the value of the source. Removing it entirely would require a more significant refactoring of the code.

jibel: Thanks for your comment. I updated the PR accordingly.

CarlosNihelton (approved):
LGTM :+1:

PR #105, opened 2023-08-25
fix: Do not show the security key page for ZFS

jibel:
This option is not yet supported in subiquity and must not be displayed.

closes: https://github.com/canonical/ubuntu-desktop-installer/issues/2311


CarlosNihelton (approved):
LGTM 👍🏽

PR #107, opened 2023-08-29
take first capability from allowed as default

mwhudson:
Before https://github.com/canonical/subiquity/pull/1774 the order of the capabilities in `allowed` was arbitrary, resulting in https://bugs.launchpad.net/ubuntu-desktop-installer/+bug/2033093 and other confusing behaviour. The subiquity PR puts the capability that should be considered the default first, so that's what the client side code should do too.


mwhudson:
I haven't tested this (in fact I made this branch by clicking around in github, not even locally)

dbungert (approved):
I have enough of dryrun working to see this working as intended.
I ran subiquity with and without the enum.value sorting, and could see it select TPM or not depending on the sort behavior, as expected.
Please merge.

PR #131, opened 2023-09-19
feat: mark network configured on ConfirmPage

d-loose:
Moves the call to the network service's `markConfigured()` from the NetworkPage to the ConfirmPage in order to resolve https://bugs.launchpad.net/subiquity/+bug/2017278 .


ogayot (approved):
I tested a build with this patch. I started the install offline and then connected using the top bar upon seeing that the checkbox was grayed out. Once connected, I ticked the checkbox and the install proceeded as expected. 

Codecs were installed to the target system as expected.

Looks good to me from the functional standpoint. Good work ! :)

PR #134, opened 2023-09-19
feat: update install option names

d-loose:
This updates the titles and subtitles of the installation options as follows:
```
1. (ubuntu-desktop-minimal): Default installation - Just the essentials, web browser and basic utilities
2. (ubuntu-desktop): Expanded installation - An offline-friendly selection of office tools, utilities, web browser and games
```

I've renamed the existing "normalInstallation*" l10n strings to "expandedInstallation*" to avoid confusion between "normal" and "default".

All previously existing translations have been removed.

|Before|After|
|-|-|
|Screenshot from 2023-09-19 16-24-46|Screenshot from 2023-09-19 16-37-28|

Screenshot created using the most recent 'sources' configuration:
```yaml
- default: true
  description:
    en: A minimal but usable Ubuntu Desktop.
  id: ubuntu-desktop-minimal
  locale_support: langpack
  name:
    en: Ubuntu Desktop (minimized)
  path: minimal.squashfs
  preinstalled_langs:
  - de
  - en
  - es
  - fr
  - it
  - pt
  - ru
  - zh
  - ''
  size: 3990839296
  type: fsimage-layered
  variant: desktop
  variations:
    minimal:
      path: minimal.squashfs
      size: 3990839296
    minimal-enhanced-secureboot:
      path: minimal.enhanced-secureboot.squashfs
      size: 4294500352
      snapd_system_label: enhanced-secureboot-desktop
- description:
    en: A full featured Ubuntu Desktop.
  id: ubuntu-desktop
  locale_support: langpack
  name:
    en: Ubuntu Desktop
  path: minimal.standard.squashfs
  preinstalled_langs:
  - de
  - en
  - es
  - fr
  - it
  - pt
  - ru
  - zh
  - ''
  size: 5712293888
  type: fsimage-layered
  variant: desktop
  variations:
    enhanced-secureboot:
      path: minimal.standard.enhanced-secureboot.squashfs
      size: 5983289344
      snapd_system_label: enhanced-secureboot-desktop
    standard:
      path: minimal.standard.squashfs
      size: 5712293888
```


tim-hm (approved):

tim-hm:
Looks great thanks for making those changes :pray:

mwhudson:
I realize this probably isn't very realistic at this point, but it would be nice if the description was taken from the source catalog somehow...

d-loose:
That'd be a trivial change from a technical point of view, since subiquity already provides us with the localized descriptions from the sources catalog. As far as I'm aware it's currently being handled in the UI to be able to control all l10n strings from a single place.

@local-optimum, @tim-hm do you have any opinions on this?

tim-hm:
My worry is if we start to pull l10n strings from subiquity we're then getting UI strings from two sources and overly complicating things down the line. My intuition is to keep these delineated meaning UDI expects data/enums back from subiquity and then relies on the dart l10n stack to update the GUI accordingly.

mwhudson:
Yeah l10n is the problem. Subiquity does support l10n of the string from the catalog but we never really figured out how to get the translated strings onto the ISO. Having the source catalog have some kind of enum/tag that the client can expand into a nice human description would be another option, it somehow seems ugly for a new kind of source to require source changes in the client though. Basically I don't know what the answer is!

PR #136, opened 2023-09-20
fix(logging): specify path to log file, not directory

d-loose:
UDENG-1423
Fix https://github.com/canonical/ubuntu-desktop-installer/issues/2342.

The `path` parameter in `Logger.setup()` from `ubuntu_logger` is supposed to be the path to the actual log file, not the directory containing it.


tim-hm (approved):
I've tested and the logs are now being written to `/var/log/installer/ubuntu_desktop_installer.log`

PR #140, opened 2023-09-21
feat(slides): update accessibility slide

d-loose:
* Lists LibreOffice Writer as 'available' instead of 'included'
* Moves Language Support and Orca Screen Reader to the first row
image


tim-hm (approved):
LGTM

PR #144, opened 2023-09-22
Allow to login in the newly created session after setup

sergio-costas:
This patch adds a new final page to ubuntu_init where the button to launch the store is removed (since it doesn't work on core desktop) and that launches the newly created session by calling GDM.

This patch requires the patches for dbus.dart that implements support for point-to-point dbus connections: https://github.com/canonical/dbus.dart/pull/363


packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
```diff
@@ -0,0 +1,57 @@
...
+  static Future<bool> load(WidgetRef ref) {
+    return ref.read(launchsessionModelProvider).init();
+  }
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final l10n = LaunchsessionLocalizations.of(context);
+    return WizardPage(
+      title: YaruWindowTitleBar(
+        title: Text(l10n.storeTitle),
```
kenvandine, line 22: I assume this shouldn't be using storeTitle?

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
```diff
@@ -0,0 +1,57 @@
...
+      ),
+      content: Column(
+        crossAxisAlignment: CrossAxisAlignment.center,
+        children: [
+          Expanded(
+            child: SvgPicture.asset('assets/store.svg', package: 'ubuntu_init'),
```
kenvandine, line 28: Also probably shouldn't use the store image

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
```diff
@@ -0,0 +1,57 @@
...
+          ),
+          Text(
+            l10n.storeHeader,
```
kenvandine, line 31: storeHeader is probably not correct

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
```diff
@@ -0,0 +1,57 @@
...
+            style: Theme.of(context).textTheme.titleLarge,
+          ),
+          FractionallySizedBox(
+            widthFactor: 0.5,
+            child: Text(l10n.storeDescription, textAlign: TextAlign.center),
```
kenvandine, line 36: same for storeDescription

kenvandine (changes requested):
The new page is using text and an image from the store page.  We need new strings for this page, I'd recommend modeling it off gnome-initial-setup's final page.

It would also be good to include a screenshot of the page in the PR

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
sergio-costas: I copied the store page and removed the button for launching the store.

sergio-costas: I'll do the changes required to model it from the final one in gnome-initial-setup.

sergio-costas: Changed.

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
sergio-costas: Fixed.

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
sergio-costas: Fixed.

sergio-costas:
Changed the launchsession page to mimic the one in gnome-initial-setup.

imagen

sergio-costas:
I don't know what is happening with github... the "git push" command doesn't work.

sergio-costas:
Uploaded.

sergio-costas:
Rebased to fix the conflicts.

packages/ubuntu_init/assets/ready-to-go.svg
d-loose, line 1: Yaru as a similar looking icon icon (`YaruIcons.ok_filled`) that you could use without the need for an svg asset.

packages/ubuntu_init/lib/src/init_wizard.dart
```diff
@@ -112,11 +114,15 @@ class InitWizard extends ConsumerWidget {
           builder: (_) => const PrivacyPage(),
           onLoad: (_) => PrivacyPage.load(ref),
         ),
-        InitRoutes.store: WizardRoute(
-          builder: (_) => const StorePage(),
-          onLoad: (_) => StorePage.load(ref),
```
d-loose, line 117: Seems like you have accidentally removed the `StorePage` here when adding the new `LaunchSessionPage` - please restore it :)

packages/ubuntu_init/lib/src/launchsession/launchsession_model.dart
```diff
@@ -0,0 +1,14 @@
...
+
+final launchsessionModelProvider = ChangeNotifierProvider(
+  (ref) => LaunchsessionModel(),
+);
+
+class LaunchsessionModel extends ChangeNotifier {
+  LaunchsessionModel();
+
+  Future<bool> init() async => true;
+}
```
d-loose, line 14: See comment above - let's create a method here that fetches the identity and calls your service's `openNewSession`.

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
```diff
@@ -0,0 +1,56 @@
...
+            FractionallySizedBox(
+              widthFactor: 0.5,
+              child: Text(l10n.launchsessionDescription, textAlign: TextAlign.center),
+            ),
+          ].withSpacing(kWizardSpacing),
+      ),
+      bottomBar: WizardBar(
+        leading: WizardButton.previous(context),
+        trailing: [
+          WizardButton.next(context),
```
d-loose, line 43: Finally, you can call the method from the model in the button's `onNext` here.

packages/ubuntu_init/lib/src/init_wizard.dart
sergio-costas: The problem is that there can't be two end pages: only the last one will show the "Done" button, while the other will show the "Next" button, and in the ubuntu-core-desktop-init program the StorePage isn't used.

But since this is common for all the wizards... how can we solve this?

packages/ubuntu_provision/lib/src/services/gdm_service.dart
d-loose, line 1: Could you run `dart format` on this file to appease the linter? :)

packages/ubuntu_init/assets/ready-to-go.svg
sergio-costas: I just used the same from gnome-initial-setup, just in case... I'll check if I can change the color... colour... I think that I have the UK english dictionary configured...

packages/ubuntu_init/lib/src/init_wizard.dart
```diff
@@ -112,11 +114,15 @@ class InitWizard extends ConsumerWidget {
...
+        InitRoutes.launchsession: WizardRoute(
+          builder: (_) => const LaunchSessionPage(),
+          onLoad: (_) => LaunchSessionPage.load(ref),
           onNext: (_) async {
             final window = YaruWindow.of(context);
+            IdentityService identityService = getService<IdentityService>();
+            var identity = await identityService.getIdentity();
+            var gdmService = GdmService();
+            await gdmService.openNewSession(identity.username, identity.password);
```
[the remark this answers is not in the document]
sergio-costas: Ok.

packages/ubuntu_provision/lib/src/services/gdm_service.dart
sergio-costas: I did, and...

```
$ dart format gdm_service.dart 
Formatted 1 file (0 changed) in 0.20 seconds.
```

packages/ubuntu_init/assets/ready-to-go.svg
d-loose: Ah I see - keep the svg if you want to be visually consistent with gnome-initial-setup, of course

packages/ubuntu_init/lib/src/init_wizard.dart
d-loose: I didn't test it, but you should be able to solve this by adding `userData: const WizardRouteData(hasNext: false)` to both `InitRoutes.store` and `InitRoutes.launchsession`.

sergio-costas: I'll test it. Thanks!

sergio-costas: I tested, and it only shows the page that is placed before in the code. If I put first the code for the store page and in the ubuntu-provision.conf file I put the launchsession page, it just exits before showing the launch session page (but the store one works fine). If I put first the code for the launch session page and in the ubuntu-provision.conf file I put the store page, when I press "Next" in the previous page, it shows nothing and calls the "onNext" code of the launch session page... (!?)

sergio-costas: It seems like the "onNext" code is called no matter if the page was shown or not...

sergio-costas: A quick&dirty solution would be to put a public flag in the pages that is set to FALSE by default, and to TRUE if the page has been shown, and only run the specific code in the "onNext" callback if the flag is TRUE... What do you think?

d-loose: Had a quick look at this. You're right, it seems like the explicitly provided `onNext` is called even when the page is not included. I don't have the time right now to think about how to properly change this in `wizard_router`.
What we could do as a fix is check again in `InitRoute.store`'s `onNext` whether the route is actually included and then act accordingly, e.g.

```dart
InitRoutes.store: WizardRoute(
  builder: (_) => const StorePage(),
  onLoad: (_) => StorePage.load(ref),
  userData: const WizardRouteData(hasNext: false),
  onNext: (settings) async {
    if (_hasRoute(settings.name!) &&
        ref.read(initModelProvider).hasRoute(settings.name!)) {
      final window = YaruWindow.of(context);
      await _onDone?.call();
      await window.close();
      return InitRoutes.initial;
    }
    return InitRoutes.launchsession;
  },
),
InitRoutes.launchsession: WizardRoute(
  builder: (_) => const LaunchSessionPage(),
  onLoad: (_) => LaunchSessionPage.load(ref),
  userData: const WizardRouteData(hasNext: false),
  onNext: (_) async {
   ...
  },
),
```

I'd have to check if this as any unwanted implications, though

sergio-costas: I'll check this and other solutions...

sergio-costas: I found a simple way:

```
InitRoutes.launchsession: () {
          bool shownPage = false;
          return WizardRoute(
            userData: const WizardRouteData(hasNext: false),
            builder: (_) => const LaunchSessionPage(),
            onLoad: (_) {
              shownPage = true;
              return LaunchSessionPage.load(ref);
            },
            onNext: (_) async {
              if (shownPage) {
                final window = YaruWindow.of(context);
                GdmService gdmService = getService<GdmService>();
                await gdmService.openNewSession();
                await _onDone?.call();
                await window.close();
                return InitRoutes.initial;
              } else {
                return null;
              }
            }
          );
        }(),
```

packages/ubuntu_init/lib/src/init_wizard.dart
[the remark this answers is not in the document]
sergio-costas: Done

packages/ubuntu_init/lib/src/launchsession/launchsession_model.dart
sergio-costas: Done

packages/ubuntu_init/lib/src/launchsession/launchsession_page.dart
sergio-costas: Done

kenvandine (approved):

sergio-costas:
@kenvandine I misunderstood the comment from Robert at his dbus.dart branch; anyway, I tested it and the code does work with the current code in main, so that branch doesn't need to be merged. Also, I had to revert a change that I did, because it is needed to "reuse" the GDM connection for the greeter and the user verifier.

sergio-costas:
@d-loose Should I close this, since it seems to have been implemented with https://github.com/canonical/ubuntu-desktop-provision/pull/298 ?

d-loose:
Yes :)
Let's discuss additional requirements for Core Desktop in Madrid if there are any!

PR #152, opened 2023-09-26
fix(ci): semi-automated autoinstall test

d-loose:
This seems to fix the recent issues with semi-automated autoinstall test, which were presumably caused by #131.

I haven't had the time to debug this locally so I can't provide any subiquity logs yet. Giving subiquity some time between marking the network configured and starting the installation resolves the current problem, but this might mask an underlying issue that we should investigate (I'd open a separate issue if we decide to accept this as a fix for now).


tim-hm:
@d-loose I don't believe this is necessary with the upstream fixes from https://github.com/canonical/ubuntu-desktop-provision/pull/155 so will close unless you object?

PR #158, opened 2023-10-02
fix(lint): remove superfluous lint rules

spydon:
Just removing some superfluous lint rules:
* `lines_longer_than_80_chars`  and `public_member_api_docs` already default to false with the rulesets you have.
* `use_super_parameters` is set to true [here](https://github.com/dart-lang/lints/blob/main/lib/recommended.yaml#L68) already.


spydon:
My default GitHub e-mail was not added to my launchpad account. I added it now, so hopefully just re-running the CLA action should work.

d-loose (approved):
Thanks, LGTM :+1:

PR #160, opened 2023-10-04
chore: Pin Flutter and SDK version to 3.10.6/3.0.6

spydon:
Since the Flutter version 3.10.6 should be used until the Ubuntu LTS version is released I suggest that we pin the version until then.


spydon:
Closing this since it is not supposed to be pinned to this version and now getting updated.

PR #165, opened 2023-10-10
Add correct end page for the Core Desktop installer

sergio-costas:
The current Ubuntu-core-desktop-init program used for the installer shows, at the end, the same page than the live installer: one where the user can choose between “reboot” or “continue testing”. This is invalid for Core Desktop because it’s not a live session, so choosing “continue testing” leaves a graphical session without desktop.

This patch adds a new parameter in whitelabel.yaml, `live-session`, which by default is true. When this parameter is set to false, it will change the last page to one that only allows to reboot, and shows a more fitting message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,9 @@ The configuration file has the following structure:
 # Options:
 # - standard (default): the common provision flow for Ubuntu Desktop and Flavours
 # - oem: enables the eula page and disables the user creation page during bootstrap
-mode: standard | oem
+# - core-desktop: disables exiting the installer at the end (because there is no
+#   live session), and only allows to reboot
+mode: standard | oem | core-desktop
 
 # (Optional) When set, the light and dark theme is inherited from ubuntu-flutter-plugins and the distro name is set.
 #
--- a/packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
@@ -17,6 +17,7 @@ final installModelProvider = ChangeNotifierProvider(
     getService<JournalService>(),
     getService<ProductService>(),
     getService<SessionService>(),
+    getService<ConfigService>(),
   ),
 );
 
@@ -60,12 +61,19 @@ class InstallationEvent {
 /// View model for [InstallPage].
 class InstallModel extends SafeChangeNotifier {
   /// Creates an instance with the given client.
-  InstallModel(this._client, this._journal, this._product, this._session);
+  InstallModel(
+    this._client,
+    this._journal,
+    this._product,
+    this._session,
+    this._config,
+  );
 
   final SubiquityClient _client;
   final JournalService _journal;
   final ProductService _product;
   final SessionService _session;
+  final ConfigService _config;
 
   Stream<String>? _log;
   ApplicationStatus? _status;
@@ -154,15 +162,19 @@ class InstallModel extends SafeChangeNotifier {
     notifyListeners();
   }
 
+  late final ProvisioningMode _provisioningMode;
+
+  ProvisioningMode get provisioningMode => _provisioningMode;
+
   /// Initializes and starts monitoring the status of the installation.
-  Future<void> init() {
-    return _client.getStatus().then((status) {
-      _log = _journal.start([status.logSyslogId, status.eventSyslogId]);
-      _events = _journal.start([status.eventSyslogId],
-          output: JournalOutput.cat).listen(_processEvent);
-      _statuses = _client.monitorStatus().listen(_updateStatus);
-      _updateStatus(status);
-    });
+  Future<void> init() async {
+    final status = await _client.getStatus();
+    _log = _journal.start([status.logSyslogId, status.eventSyslogId]);
+    _events = _journal.start([status.eventSyslogId],
+        output: JournalOutput.cat).listen(_processEvent);
+    _statuses = _client.monitorStatus().listen(_updateStatus);
+    _updateStatus(status);
+    _provisioningMode = await _config.provisioningMode;
   }
 
   @override
--- a/packages/ubuntu_bootstrap/lib/pages/install/install_page.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/install/install_page.dart
@@ -188,6 +188,9 @@ class _DonePage extends ConsumerWidget {
     final flavor = ref.watch(flavorProvider);
     final lang = UbuntuBootstrapLocalizations.of(context);
     final model = ref.watch(installModelProvider);
+    final isCoreDesktop =
+        model.provisioningMode == ProvisioningMode.coreDesktop;
+
     return WizardPage(
       headerPadding: EdgeInsets.zero,
       contentPadding: EdgeInsets.zero,
@@ -206,13 +209,17 @@ class _DonePage extends ConsumerWidget {
               crossAxisAlignment: CrossAxisAlignment.stretch,
               children: [
                 MarkdownBody(
-                  data: lang.readyToUse(model.productInfo),
+                  data: isCoreDesktop
+                      ? lang.rebootToConfigure(model.productInfo)
+                      : lang.readyToUse(model.productInfo),
                   styleSheet: MarkdownStyleSheet(
                     p: Theme.of(context).textTheme.titleLarge,
                   ),
                 ),
                 const SizedBox(height: kWizardSpacing * 1.5),
-                Text(lang.restartWarning(flavor.displayName)),
+                Text(isCoreDesktop
+                    ? lang.rebootToConfigureWarning
+                    : lang.restartWarning(flavor.displayName)),
                 const SizedBox(height: kWizardSpacing * 1.5),
                 Row(
                   children: [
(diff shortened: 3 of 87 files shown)


sergio-costas:
Just rebased this to allow it to be merged...

sergio-costas:
Ok!

tim-hm:
@d-loose read my mind! Yes we should fold into white label.yaml

sergio-costas:
Ok... something like this?

packages/ubuntu_bootstrap/lib/pages/install/install_page.dart
```diff
@@ -6,22 +7,31 @@ import 'package:ubuntu_bootstrap/l10n.dart';
...
 import 'package:ubuntu_widgets/ubuntu_widgets.dart';
 import 'package:ubuntu_wizard/ubuntu_wizard.dart';
 import 'package:yaru_icons/yaru_icons.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
 class InstallPage extends ConsumerWidget with ProvisioningPage {
-  const InstallPage({super.key});
+  InstallPage({super.key});
+
+  bool liveSession = true;
```
spydon, line 20: This should be stored in the `installModelProvider`, since you shouldn't have mutable state here.

sergio-costas: Ok, I think that it is done... I still have to test it, anyway.

spydon: Looks correct now, good job!

README.md
```diff
@@ -19,6 +19,8 @@ The Flutter UI can be configured using a [YAML](https://yaml.org/) file.
     elevated-button-text-color: "#0000ff"
     window-title: "Custom Window Title"
 
+  live-session: true
```
d-loose, line 22: Do we want a completely separate config key for this, or should it be part of the `mode` enum?

sergio-costas: That's your decision... I'll put it wherever you prefer.

d-loose: In any case - could you rebase and merge the latest readme update @spydon landed in #431? I also think we should sync any changes to the whitelabel part in the readme with the [configuration guide](https://docs.google.com/document/d/10R0YOj4e8BTv6XPw9OE_y7GDy72xPqA5XP5lu0M7VbE/edit#heading=h.x8jmsnxxm1fh).

spydon: I think this would fit well in the `mode` enum, we'll never have OEM core installations right?

d-loose: Well, I wouldn't say *never*, but I guess in that case we could add an `oem-core` mode :D

packages/subiquity_client/subiquity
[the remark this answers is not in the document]
sergio-costas: Ops... yes

README.md
sergio-costas: Rebased and squashed (it wasn't trivial to rebase without it)

sergio-costas: Mmm... now I see what you mean, and I'm not sure, because using "oem-core" presumes that the only possibility for core desktop is to be distributed in a disk image without a live environment... And also that only core desktop will be distributed in a non-live disk image...

d-loose: Does core desktop expect user creation in stage 3, or is that supposed to be done later during first boot in stage 4? In the latter case we'd need to treat the user creation the same way we do in the oem case anyways.
In general, I wouldn't worry about possible other "modes" just yet. For the time being we care about:
* `standard` (Ubuntu Desktop + Flavors, live iso with user creation in stage 3 and no stage 4 at all)
* `oem` (additional eula page, no user creation in stage 3, init wizard with user creation on first boot aka stage 4)
* `core-desktop` (installation from image, no live session, ....?)

sergio-costas: Ok.

sergio-costas: Done.

sergio-costas:
@spydon I rebased, but there are two tests that don't pass, and I'm unable to discover why. Can you help me?

spydon:
> @spydon I rebased, but there are two tests that don't pass, and I'm unable to discover why. Can you help me?

Run `melos generate` and it will generate the mocks :)

sergio-costas:
Next time I'll read first the "CONTRIBUTE.md" file 😅

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
```diff
@@ -186,6 +193,8 @@ class InstallModel extends SafeChangeNotifier {
         await precacheImage(AssetImage(asset), context);
       }
     }
+    final value = await _config.provisioningMode;
+    _provisioningMode = value;
```
d-loose, line 197: Could we move this into `init()` and turn it into a one-liner?

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
```diff
@@ -172,6 +175,10 @@ class InstallModel extends SafeChangeNotifier {
     super.dispose();
   }
 
+  var _provisioningMode = ProvisioningMode.standard;
```
d-loose, line 178: This might be one of the cases where `late final` might actually be better (I'd consider a runtime error 'better' than returning the wrong provisioning mode here). 
@spydon any thoughts?

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
sergio-costas: Something like this...?

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
sergio-costas: Changed to late final. Opinions?

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
```diff
@@ -60,12 +61,14 @@ class InstallationEvent {
 /// View model for [InstallPage].
 class InstallModel extends SafeChangeNotifier {
   /// Creates an instance with the given client.
-  InstallModel(this._client, this._journal, this._product, this._session);
+  InstallModel(
+      this._client, this._journal, this._product, this._session, this._config);
```
spydon, line 65: Add a trailing comma after `this._config` and reformat

spydon (commented):
Lgtm, just one minor comment.
And I think that there are some formatting issues, they should be solved if you run `melos format`.

packages/ubuntu_bootstrap/test/install/install_model_test.dart
```diff
@@ -134,10 +133,9 @@ void main() async {
 
     final session = MockSessionService();
     final config = MockConfigService();
+    when(config.provisioningMode).thenReturn(Future<ProvisioningMode>.value(ProvisioningMode.standard));
```
d-loose: Could you use `thenAnswer` here? See [docs](https://pub.dev/packages/mockito#a-quick-word-on-async-stubbing).

d-loose: You don't need a `Future.value` then, but instead use
```dart
when(config.provisioningMode).thenAnswer((_) async => ProvisioningMode.standard));
```

sergio-costas: Good point! Changed.

sergio-costas:
I'm trying to use `melos test`, but I receive several errors. Specifically:

/home/raster/workspace/ubuntu-desktop-provision/packages/subiquity_client/test/subiquity_client_test.dart:
- `Shell: TypeError: expected string or bytes-like object, got 'NoneType'` (when a test tries to compile a regular expression)
- `SocketException: Connection failed (OS Error: No such file or directory, errno = 2), address = subiquity/.subiquity/socket, port = 0` (before it was that subiquity/.subiquity path didn't exist, so I manually created the folder, but now it returns this)

Any ideas?

d-loose:
Since this particular test passes in the CI, I suppose it's a local issue on your side. Make sure your repository is clean and update the subiquity git submodule via `git submodule update --init`

packages/ubuntu_bootstrap/lib/pages/install/install_model.dart
spydon:
```suggestion
  InstallModel(
    this._client,
    this._journal,
    this._product,
    this._session,
    this._config,
  );
```

sergio-costas: Mmm... It's already changed...

sergio-costas: Yes, yes... sorry. That answer was for the "lgtm" part. Maybe I should put this as a draft until I add that test. Sorry.

sergio-costas:
Ok, my problem with `melos test` isn't strictly speaking in my system, but in `python3-pyroute2`... It seems that the .deb is broken!

sergio-costas:
Fixed the problem with python3-pyroute2. It was another module that I installed from GIT, that had wrong permissions in the EGG folder.

sergio-costas:
@d-loose All ready. Added a test for core desktop too.

packages/ubuntu_provision/test/services/page_config_service_test.dart
```diff
@@ -38,13 +37,16 @@ pages:
...
 MockConfigService createMockConfigService({
   String config = '',
-  ProvisioningMode mode = ProvisioningMode.standard,
+  ProvisioningMode? mode,
 }) {
   final mock = MockConfigService();
   final yaml = loadYaml(config)?['pages'] as YamlMap?;
-  when(mock.get('mode')).thenAnswer((_) async => mode.name);
+  when(mock.get('mode')).thenAnswer((_) async =>
+      mode != null ? mode.name : ProvisioningMode.standard.toString());
```
d-loose, line 45: One small detail:
```suggestion
      mode != null ? mode.name : ProvisioningMode.standard.name);
```
I think `toString()` would give `"ProvisioningMode.standard"` instead of `"standard"`

d-loose (approved):
Awesome, LGTM :+1:

d-loose:
Some cleanup would definitely be nice :+1: I don't have strong opinions about squashing, so that's fine with me too :)

spydon (approved):
Lgtm! Probably the easiest to squash all the commits :)

PR #174, opened 2023-10-14
[ ubuntu_bootstrap_uk.arb ] Add Ukrainian l10n🇺🇦

GoldenVadim:
nice translation?


taurus-forever (approved):
As a Ukrainian speaker, I confirm that nothing rude or irrelevant here.

PR #176, opened 2023-10-15
Update ubuntu_init_localizations_ru.dart

xuserz:
fix translate localization


spydon:
The new weblate instance is now up and all translations are done through there.

PR #179, opened 2023-10-19
feat: update to latest flutter and yaru versions, use asdf for version management

d-loose:
After bumping the package versions in [ubuntu-flutter-plugins](https://github.com/canonical/ubuntu-flutter-plugins/compare/97d48acc9cef1e427630a8fc1d7a6ddfac811a76...5eb08476e9d3116c4a76a07d83d353b60284a8c1) and [yaru_test](https://github.com/ubuntu/yaru_test.dart/pull/31), we can now upgrade to the latest version of flutter and yaru here.

@elioqoshi, I noticed that you had a look at the [design changes in yaru.dart](https://github.com/ubuntu/yaru.dart/compare/v0.9.0...v1.1.0) and approved them. With this PR we'd inherit those here.

For reference, [here](https://github.com/ubuntu/yaru_widgets.dart/compare/v2.5.0...v3.2.0) are the changes in yaru_widgets, which are mostly of technical nature, and [here](https://github.com/ubuntu/yaru_icons.dart/compare/1.0.0...v2.2.1) are the changes in yaru_icons (new icons and technical improvements).

@tim-hm I added `.tool-versions` for use with [asdf](https://github.com/asdf-vm/asdf) and updated the CI accordingly.


tim-hm:
Interesting, I hadn't thought about using asdf directly in CI 🤔 I was thinking primarily about managing tooling versions locally. 

@didrocks do you have any experience using a version manager in CI / concerns on this approach?

didrocks:
> @didrocks do you have any experience using a version manager in CI / concerns on this approach?

In python mostly, using `virtualenv`. This is built-in in Go, so we don’t really have that issue.

My main concern is, whatever solutions we pick, we need to take this into account:
- an easy way for everyone contributing to the project to ensure that they are using the exact same compiler toolchain/stdlib. (This should be documented in CONTRIBUTING.md)
- ensuring that the CI is picking the exact same revision than the one selected above (to avoid "pass locally but not on CI, oh, this is because the flutter/dart version is different")
- and finally, ensuring that the finale built artifact (snaps or debs) is also using the exact same version. This is easier for snap, given the provided toolchain selection is correct.

So, as long as the toolchain selection process allows to define a version only in one place, and this is propagated on those 3 different flows, I’m fine with it.

d-loose:
In order to build a snap using the same flutter version without explicitly specifying it in multiple places, we could do something like this: https://github.com/canonical/firmware-updater/pull/218/commits/cf9e3103a5e79b23399f3ae8f2ec829342c542f9 (tested [here](https://github.com/d-loose/firmware-updater/actions/runs/6588768244/job/17901781765))

tim-hm:
Lets give it a try! We can adjust easily enough if its awkward.

d-loose:
@tim-hm considering https://github.com/canonical/firmware-updater/pull/218#discussion_r1368254740: we'd have to fix (and maintain) the flutter version in seven different pubspec.yamls here - is this what we want to do?

tim-hm:
@d-loose any chance the package versions can be inherited from the root pubspec.yaml

d-loose:
> @d-loose any chance the package versions can be inherited from the root pubspec.yaml

As far as I can tell this isn't possible, since each package is still an independent entity. `melos bootstrap` creates custom `pubspec_overrides.yaml` files to link together the packages it manages, but doesn't provide an option to specify common environment constraints.

d-loose:
@tim-hm after encountering some weird behavior locally I also found [this issue](https://github.com/flutter/flutter/issues/95472) - it seems like flutter only respects lower bounds on the version, while upper bounds and exact pins don't work.
I'd suggest leaving the constraints as they are in this PR. I'm currently working on updating the CONTRIBUTING.md and would aim to clarify it there. Automated tooling like the CI or snapcraft.yamls should use the version specified in `.tool-versions` anyways.

tim-hm:
I think we should keep the specific version rather than expanding as this PR does.

Firstly, a fixed version signals to contributors that we're expecting them to run the codebase using that specific version of flutter -- regardless if the tooling enforces it or not. We've seen it break between point releases (ie x.z.6 worked but x.z.7 didn't) and this is a standard way to signal the expected version.

Secondly, we should assume that specific versions will be respected by the tooling rather than expanding the range to accomodate what is likely a bug upstream.

@didrocks any thoughts?

didrocks:
> Firstly, a fixed version signals to contributors that we're expecting them to run the codebase using that specific version of flutter -- regardless if the tooling enforces it or not. We've seen it break between point releases (ie x.z.6 worked but x.z.7 didn't) and this is a standard way to signal the expected version.

> Secondly, we should assume that specific versions will be respected by the tooling as it's likely a bug upstream.

I don’t really understand the 2 points, are they seem to contradicts each other, no? The second one sounded to me "we should listen and trust the tooling when following semantic versioning", or is it that we should trust them to not retag with the same version some tools, or that the tooling will respect our fixed version?

Did we see a regular breaking pattern when using upstream version or was it a one of (and did we detect it when building or with tests?). I’m a fan on MVS as a pattern, accepted to trust it with time. Is there a way to separate version requirements and selected toolchains in Dart/Flutter?

If the breaking pattern is regular or if there is no way to pin the toolchain independently of the requirements, I recommend then enforcing a fixed version.

spydon:
With the new Melos 3.2.0 feature we can specify common Flutter dependency constraints in the `melos.yaml` file, so that we don't have to manually update them on so many places. I can set up a PR pointing towards this one so that you can see how it would look like.
https://melos.invertase.dev/commands/bootstrap#shared-dependencies

PR #182, opened 2023-10-25
feat: update CONTRIBUTING.md

d-loose:
Updates the contribution guidelines, using [our template](https://github.com/canonical/desktop-engineering/blob/main/project-repo/templates/common/CONTRIBUTING.md) as a starting point.
I've left out the issue reporting section, since we'll probably need to rethink that process first (currently there's [launchpad](https://bugs.launchpad.net/ubuntu-desktop-installer/), the issue tracker at [ubuntu-desktop-installer](https://github.com/canonical/ubuntu-desktop-installer/issues) and the one in this repo [here](https://github.com/canonical/ubuntu-desktop-provision/issues)).
I've also left out the section on how to build/run the binaries for now. The two binaries that can currently be built are `ubuntu_bootstrap` and `ubuntu_init`, while the rest of the packages are pure libraries. Maybe it's a good idea to separate those from each other similar to the typical go project layout the new backend will use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,125 @@
-# Contribution Guide
+# Contributing to Ubuntu Desktop Provision
 
-If you have a problem, please [file an issue](https://github.com/canonical/ubuntu-desktop-provision/issues/new).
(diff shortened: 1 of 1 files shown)


matthew-hagemann (approved):
LGTM!

PR #186, opened 2023-10-27
Add flavo(u)r data in ProductService

sergio-costas:
It can be useful to have available the flavor data passed during the call to runInstallerApp (for example, to change the page appearance for specific Ubuntu flavors, like Ubuntu Core Desktop, based on the id field).

This MR makes that data available in the ProductService class, thus allowing to retrieve it anywhere.

### DIFF
--- a/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
+++ b/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
@@ -437,7 +437,8 @@ class FakeProductService implements ProductService {
   ProductInfo getProductInfo() => ProductInfo(name: 'Ubuntu', version: '23.04');
 
   @override
-  UbuntuFlavor? getFlavor() => const UbuntuFlavor(id: 'FakeUbuntu', name: 'Ubuntu Fake Flavo(u)r');
+  UbuntuFlavor? getFlavor() =>
+      const UbuntuFlavor(id: 'FakeUbuntu', name: 'Ubuntu Fake Flavo(u)r');
 
   @override
   String getReleaseNotesURL(String languageCode) =>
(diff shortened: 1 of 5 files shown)


d-loose:
I'm wondering why it's necessary to add this to the `ProductService` - is there a situation where accessing the flavor through the `flavorProvider` doesn't work?

sergio-costas:
My idea is to be able to get the value passed by the installer when it calls the runAppInstaller() function, because I suspect that it can be different than the autodetected one. The idea is to allow to run specific customizations in the pages for specific flavours (like core desktop).

Of course, if the autodetection is flawless, or flavorProvider already returns that value, then we can close this.

Also, I can move this into flavorProvider, if it makes more sense.

d-loose:
> My idea is to be able to get the value passed by the installer when it calls the runAppInstaller() function, because I suspect that it can be different than the autodetected one.

If you explicitly specify a flavor in `runInstallerApp()` (like the core desktop installer does), the `flavorProvider` will be overridden with that value (see [here](https://github.com/canonical/ubuntu-desktop-provision/blob/318a8e8a393f49dfa76392abe52f73870d96a5d8/packages/ubuntu_bootstrap/lib/installer.dart#L149-L153)), so you don't need to rely on `UbuntuFlavor.detect()` at all.

> The idea is to allow to run specific customizations in the pages for specific flavours (like core desktop).

You should be able to do this without any changes. A lot of pages use the `flavorProvider` simply to pass the flavor name into an l10n string (see e.g. [here](https://github.com/canonical/ubuntu-desktop-provision/blob/318a8e8a393f49dfa76392abe52f73870d96a5d8/packages/ubuntu_bootstrap/lib/pages/welcome/welcome_page.dart#L31-L58)), but you could take it a step further and customize any widget explicitly:
```dart
final flavor = ref.watch(flavorProvider);

if (flavor.id == 'ubuntu-core-desktop') {
...
}
```

sergio-costas:
> If you explicitly specify a flavor in `runInstallerApp()` (like the core desktop installer does), the `flavorProvider` will be overridden with that value (see [here](https://github.com/canonical/ubuntu-desktop-provision/blob/318a8e8a393f49dfa76392abe52f73870d96a5d8/packages/ubuntu_bootstrap/lib/installer.dart#L149-L153)), so you don't need to rely on `UbuntuFlavor.detect()` at all.

Great! Then we can close this.

PR #187, opened 2023-10-27
chore: re-integrate ubuntu_desktop_installer into the repo

d-loose:
Reintegrates the ubuntu_desktop_installer package, currently maintained [here](https://github.com/canonical/ubuntu-desktop-installer) into the repo.
While we could build the installer snap directly from ubuntu_bootstrap I think it still makes sense to keep them separate until we have a better overview of the new provisioning flow.

Once this is merged, I'd add the `snap/` and `scripts/` directories from u-d-i to a separate `ubuntu-desktop-installer-snap` branch that can be used to build the installer snap.

### DIFF
--- a/packages/ubuntu_desktop_installer/.gitignore
+++ b/packages/ubuntu_desktop_installer/.gitignore
@@ -1,0 +1,53 @@
+# Miscellaneous
+*.class
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
+
+# IntelliJ related
+*.iml
+*.ipr
+*.iws
+.idea/
+
+# The .vscode folder contains launch configuration and tasks you configure in
+# VS Code which you may wish to be included in version control, so this line
+# is commented out by default.
+#.vscode/
+
+# Flutter/Dart/Pub related
+**/doc/api/
+**/ios/Flutter/.last_build_id
+.dart_tool/
+.flutter-plugins
+.flutter-plugins-dependencies
+*.lock
+.packages
+.pub-cache/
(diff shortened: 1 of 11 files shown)


matthew-hagemann (approved):
All looks good :+1:  Out of interest, I see the snap needs the `ubuntu-wsl-setup` package. Do you plan on moving that into this repo aswell?

d-loose:
I don't know why it's even a part in the snapcraft.yaml, it doesn't do anything and I'm considering to simply remove it from there.

@CarlosNihelton how's the `ubuntu_wsl_setup` package being used currently? Does it need to be part of the installer snap? We should probably discuss how to integrate this in the new provisioning flow in Riga :)

CarlosNihelton:
> I don't know why it's even a part in the snapcraft.yaml, it doesn't do anything and I'm considering to simply remove it from there.

I'm not quite following what exactly are you guys doing here in this reintegration PR. It looks like you're creating a new UDI package in here and will just bring the existing code, without its history. If I'm right, I'd be skeptical about the motivation for losing the history.

If the existing UDI repository will stay alive where it is, I'd recommend you not to bring ubuntu_wsl_setup here.

> @CarlosNihelton how's the `ubuntu_wsl_setup` package being used currently? 
That package builds the Ubuntu WSL Out of the Box experience, a GUI that onboards new users as part of the installation process. We will decommission it this cycle, so it won't live enough to justify the migration.

d-loose:
Closing this, since instead of having individual snaps for each use case, we're now aiming to build a single "ubuntu-boostrap" snap that will be configurable through an external config file.

PR #190, opened 2023-11-06
chore: Bump and introduce common dependencies

spydon:
This PR does a few (related) things:
 * It introduces common dependencies specified in `melos.yaml` so that dependency versions don't have to be bumped in every package.
 * It updates the github action so that it uses `melos bootstrap` instead of pub getting the local packages as path dependencies.
 * It removes the `pubspec.lock` file in the root of the monorepo, since it seemingly doesn't serve a purpose?
 * It removes the `analysis_options.yaml` file from `subiquity_client` and makes it use `flutter_lints` instead of `lints`, like the rest of the packages.
 * Fixes some small analyze issues that the new analyzer version found.


d-loose:
Thanks, looks really good!

Just a few questions:

> It introduces common dependencies specified in `melos.yaml` so that dependency versions don't have to be bumped in every package.

As far as I understand from reading the docs on that feature running `melos bootstrap` will then propagate the version constraints to all packages and update their respective pubspec.yamls, right? In that case, if we want to continue using dependabot to keep track of package updates, we should probably tell it to only manage the dependencies in the melos.yaml (if it is able to do that).

> It updates the github action so that it uses `melos bootstrap` instead of pub getting the local packages as path dependencies.

:+1:

> It removes the `pubspec.lock` file in the root of the monorepo, since it seemingly doesn't serve a purpose?

It was introduced in https://github.com/canonical/ubuntu-desktop-provision/commit/1447fe61642e9712cbbbb4d68855cb21305ee141 as part of the [migration to melos 3](https://melos.invertase.dev/guides/migrations#200-to-300). Is it still needed?

> It removes the `analysis_options.yaml` file from `subiquity_client` and makes it use `flutter_lints` instead of `lints`, like the rest of the packages.

We've been using the `lints` package in all of the Flutter-independent Dart libraries. Are there any benifits/drawbacks of using `flutter_lints` vs `lints`?

spydon:
> if we want to continue using dependabot to keep track of package updates, we should probably tell it to only manage the dependencies in the melos.yaml (if it is able to do that).

I don't think dependabot is able to do that currently, but it should be trivial to add that functionality to it. I'll see if I can whip up a PR to that project that does that.

> It was introduced in [1447fe6](https://github.com/canonical/ubuntu-desktop-provision/commit/1447fe61642e9712cbbbb4d68855cb21305ee141) as part of the [migration to melos 3](https://melos.invertase.dev/guides/migrations#200-to-300). Is it still needed?

Ah right, I guess it is to ensure that the CI runs the same version as locally. I think it would very rarely matter, but I'll revert the removal of the lockfile.

> We've been using the `lints` package in all of the Flutter-independent Dart libraries. Are there any benifits/drawbacks of using `flutter_lints` vs `lints`?

`flutter_lints` is just an extension of `lints`, I usually use the same rules in all projects within an organization for consistency.
Some of the flutter rules won't affect the dart project though.

d-loose:
> > if we want to continue using dependabot to keep track of package updates, we should probably tell it to only manage the dependencies in the melos.yaml (if it is able to do that).
> 
> I don't think dependabot is able to do that currently, but it should be trivial to add that functionality to it. I'll see if I can whip up a PR to that project that does that.

:+1:
 
> > It was introduced in [1447fe6](https://github.com/canonical/ubuntu-desktop-provision/commit/1447fe61642e9712cbbbb4d68855cb21305ee141) as part of the [migration to melos 3](https://melos.invertase.dev/guides/migrations#200-to-300). Is it still needed?
> 
> Ah right, I guess it is to ensure that the CI runs the same version as locally. I think it would very rarely matter, but I'll revert the removal of the lockfile.

:+1:
 
> > We've been using the `lints` package in all of the Flutter-independent Dart libraries. Are there any benifits/drawbacks of using `flutter_lints` vs `lints`?
> 
> `flutter_lints` is just an extension of `lints`, I usually use the same rules in all projects within an organization for consistency, but maybe you already have `lints` for all Dart projects, and `flutter_lints` for all Flutter projects and want to keep it that way? Some of the flutter rules won't affect the dart projects.

Ah, thanks for clarifying - it probably makes sense to use `flutter_lints` then, since subiquity_client is now part of the monorepo again.

d-loose (approved):
LGTM :+1:

PR #191, opened 2023-11-07
ci(provd): adding workflow for provd pr's

matthew-hagemann:
- Add tools module for ci jobs in provd/tools

Adding jobs for:
- go-sanity checks
- go-tests


.github/workflows/ci.yml
```diff
@@ -170,3 +177,57 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "./provd/go.mod"
+
+      - name: Install dependencies
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt update
```
didrocks, line 216: we don’t need `DEBIAN_FRONTEND=noninteractive` on apt update (and yes, I’m sure this is a copy from another place which could be fixed :p)

provd/go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/canonical/ubuntu-desktop-provision/provd
 
-go 1.21.3
+go 1.21.4
```
didrocks, line 3: is that needed? Are we vulnerable to some security issue fixed in `.4`?

provd/tools/tools.go
```diff
@@ -0,0 +1,9 @@
+//go:build tools
```
didrocks, line 1: ok, I think another PR will introduce dependabot to get those autoupdated too?

.github/workflows/ci.yml
```diff
@@ -9,6 +9,13 @@ on:
...
 
+env:
+  APT_DEPS: >-
+    libglib2.0-dev
+    gcc
+    gettext
+    libpam0g-dev
+
 jobs:
   analyze:
```
didrocks, line 20: as you have some parts specific to Go, this is specific to flutter/dart and should be marked as such?

.github/workflows/ci.yml
```diff
@@ -9,6 +9,13 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  APT_DEPS: >-
+    libglib2.0-dev
+    gcc
+    gettext
+    libpam0g-dev
```
didrocks, line 17: hum, are you sure this is needed for this project? The pam dep in particular sound like a copy from other projects where we need it :)

didrocks (changes requested):
A minor couple of comments to improve the discoverability and reduce many some undue copy/paste, but otherwise, the whole workflow sounds aligned with the others we are running. Well done!

provd/go.mod
matthew-hagemann: Yes, they were caught when I ran the sanity check the first time. Dependabot also catches them [here](https://github.com/canonical/ubuntu-desktop-provision/security/dependabot)

provd/tools/tools.go
matthew-hagemann: Yes, I will create a pr for updating the `dependabot.yaml` :+1:  these do seem to be currently out of its scope

.github/workflows/ci.yml
matthew-hagemann: Yes. In my mind it makes sense to mark `flutter-tests`, `flutter-analyze`, `flutter-coverage`, `flutter-format` & `flutter-mocks`?

matthew-hagemann:
Rebased off of main to get flutter fixes failing the pipeline

didrocks (approved):
Good to me!

As a side-note: I think a followup PR would be needed to tweak codecov. I don’t think we should fail a workflow on coverage decrease, but just showing it (btw, the fact that the coverage is changing needs to be inspected, is it not stable?).

Otherwise, we will get used to CI failing and won’t look at it. The end message that codecov is printing should be enough to look/investigate and see if the decrease of code coverage is expected or not.

.github/workflows/ci.yml
```diff
@@ -9,6 +9,14 @@ on:
   pull_request:
```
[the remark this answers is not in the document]
spydon: I think it is called `ci.yaml` in all of the flutter repositories? Not that it matters much, but that is a way more common name in GitHub actions in general too.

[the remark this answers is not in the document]
didrocks: I don’t really care about the name, we got some back and forth on CI vs QA in the past. As long as we align all our repo (most of Ubuntu repos have qa.yaml)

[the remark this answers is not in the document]
spydon: In the Ubuntu repos there are 22 called *ci.yaml and 9 called *qa.yaml, and in the Canonical repos there are 
166 called *ci.yaml and 0 called *qa.yaml, so I think that it would be easier to align by naming the `qa` ones `ci` instead? That would also align more with the common standard on GitHub.

https://github.com/search?q=org%3Aubuntu+path%3A**%2F*ci.y*ml&type=code
https://github.com/search?q=org%3Aubuntu+path%3A**%2F*qa.y*ml&type=code
https://github.com/search?q=org%3Acanonical+path%3A**%2F*ci.y*ml&type=code
https://github.com/search?q=org%3Acanonical+path%3A**%2F*qa.y*ml&type=code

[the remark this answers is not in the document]
didrocks: I’m fine with it (just a note: in reality, the actual number differs due to non public repos, but that doesn’t change much the numbers).

The only thing to take into account, is that some repos have multiple workflows being part of "CI" technically, but needed to be split for clarity, so not everything can be renamed that easily, but as long as we standardize (I think those could have some domains like `ci-<domain>` for instance. Let’s go for `ci.yaml` then and calling those the ones running on PR only!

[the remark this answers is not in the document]
didrocks: The real numbers on public ubuntu repos are: 8 `ci.y*ml` in github ubuntu org vs 9 `qa.y*ml`. But anyway, as told, let’s go with `ci.yaml`.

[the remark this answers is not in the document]
spydon: Maybe that is a pretty good idea, to split it up to `ci-flutter.yaml` and `ci-golang.yaml`?

d-loose:
> As a side-note: I think a followup PR would be needed to tweak codecov. I don’t think we should fail a workflow on coverage decrease, but just showing it (btw, the fact that the coverage is changing needs to be inspected, is it not stable?).
> 
> Otherwise, we will get used to CI failing and won’t look at it. The end message that codecov is printing should be enough to look/investigate and see if the decrease of code coverage is expected or not.

Definitely a good idea - we often get a tiny coverage decreases with small UI refactors.

PR #181, opened 2023-10-23
chore: re-integrate subiquity_client

d-loose:
* Removes the subiquity_client submodule
* (re-)imports subiquity_client and subiquity_test into the repo
* Includes subiquity as a submodule for subiquity_client (we don't have a better solution for this now, since we're generating the dart classes for the subiquity types automatically from subiquity's source)
* Includes subiquity_client and subiquity_test in the CI (essentially reverts https://github.com/canonical/ubuntu-desktop-provision/pull/34 and https://github.com/canonical/ubuntu-desktop-installer/pull/1971). We can think about optimizing the CI and running jobs in parallel after including the new backend.

PR #210, opened 2023-11-20
ci(dependabot): update to support go subproject

### LINKED ISSUE #208: Update dependabot to cover Go dependencies

matthew-hagemann:
Adding clauses to the `dependabot.yaml` to support checking go & gh-action deps

Closes #208

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -31,10 +31,15 @@ updates:
     commit-message:
       prefix: "deps(go-tools)"
 
-  ## Git Submodules
+  ## Subiquity dependencies
   - package-ecosystem: gitsubmodule
     directory: /
     schedule:
         interval: "daily"
         time: "02:00"
     target-branch: "main"
+    commit-message:
+      prefix: "deps(subiquity)"
+    allow:
+      - dependency-name: packages/subiquity_client/subiquity
+        dependency-type: direct:production
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,5 +1,37 @@
 version: 2
 updates:
+  # Infrastructure
+  ## GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+        interval: "daily"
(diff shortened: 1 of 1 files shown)


matthew-hagemann:
@d-loose I have added commit message defaults for the dependency bumps on the go deps. Is there something similar we could use for the sub-modules, or does it check more than just flutter deps?

d-loose:
The only submodule we're left with is subiquity. So I guess we could make sure dependabot only touches the subiquity submodule (in case we ever add others and forget about dependabot) and tag the commits with `deps(subiquity)`

.github/dependabot.yaml
```diff
@@ -31,10 +31,15 @@ updates:
...
   - package-ecosystem: gitsubmodule
     directory: /
     schedule:
         interval: "daily"
         time: "02:00"
     target-branch: "main"
+    commit-message:
+      prefix: "deps(subiquity)"
+    allow:
+      - dependency-name: packages/subiquity_client
```
d-loose, line 44: Not sure how `dependency-name` works here, but I believe it should either be "subiquity" (the submodule's name) or "packages/subiquity_client/subquity" (the path to the submodule)

matthew-hagemann: I based it off of one of the [commits](https://github.com/canonical/ubuntu-desktop-provision/commit/8da5726a211c3d7dfee05d1ba1ae8f925064a7bf) from dependabot in the past?

d-loose: That's from before #181 though - that submodule doesn't exist anymore. `packages/subiquity_client` is a proper directory now, but it contains subiquity as a submodule in `packages/subiquity_client/subiquity`.

matthew-hagemann: Ah apologies! I am following now. Perfect, make that change :smile:

d-loose: No worries, the naming and submodule structure can be *slightly* confusing :D

d-loose (approved):
:+1:

PR #211, opened 2023-11-20
fix(ci): fixing dependency type

matthew-hagemann:
Looking at the options for the [allow](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow) clause, production makes the most sense to me as the dependencies to keep monitored, seeing as we aren't directly working on `subiquity` in this monorepo.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -42,4 +42,4 @@ updates:
       prefix: "deps(subiquity)"
     allow:
       - dependency-name: packages/subiquity_client/subiquity
-        dependency-type: direct:production
+        dependency-type: direct


matthew-hagemann:
> > Looking at the options for the [allow](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow) clause, production makes the most sense to me as the dependencies to keep monitored, seeing as we aren't directly working on `subiquity` in this monorepo.
> 
> According to the table at the bottom of the paragraph `production` is only supported by a handful of package managers, and 'git submodule' is not one of them - so our only options would be `direct` or `all`.

Golly, sorry, didn't realize gitsubmodule was considered a pm! I was looking at what supported pip. In that case I lean to direct and only watching the explicitly defined deps. Your thoughts?

d-loose (approved):
LGTM :+1:

PR #212, opened 2023-11-20
feat: Remove post install pages

### LINKED ISSUE #199: Remove "post install" pages from ubuntu_bootstrap
In the new provisioning flow user creation, active directory configuration, setting timezone and theme are supposed to be done on first boot rather than during the installation process, so we should remove those pages from `ubuntu_bootstrap`.

d-loose:
Removes the post install pages ("identity", "active_directory", "timezone" and "theme") and their subiquity-based services from ubuntu_bootstrap.

UDENG-1755
Fix #199


packages/ubuntu_bootstrap/lib/services/installer_service.dart
```diff
@@ -16,13 +16,16 @@ class InstallerService {
...
+      // Use the default values for a number of endpoints that aren't used in
+      // the bootstrap stage, or for which a UI page isn't implemented yet.
       await _client.markConfigured([
+        'active_directory',
+        'identity',
         'mirror',
         'proxy',
         'ssh',
         'snaplist',
+        'timezone',
```
d-loose, line 28: Is it enough to mark those models as configured? I suspect we might need to adjust the `POSTINSTALL_MODEL_NAMES` in subiquity among other things since subiquity seems to create a default 'ubuntu' user if the identity model is left untouched.
@dbungert @ogayot

d-loose: Had a chat with @ogayot - right now there's no way for the UI to tell subiquity to not create a default user. This will need some changes in subiquity and possibly cloud-init.

d-loose: Turns out cloud-init already supports that, so we'll only need a small change in subiquity

tim-hm (approved):
Yes this is exactly what we're looking for 🙏

matthew-hagemann (approved):
Given it a run, given it a read. LGTM :rocket:

PR #227, opened 2023-11-22
Provd draft protos

### LINKED ISSUE #197: gRPC protofiles for provd
We need an inital draft for the protofiles for the provisioning backend. As this is more or less a 1-1 replacement for GIS, the best thing is to use it as a guide for what endpoints are needed, what fields they require and what dbus calls will need to be made.

matthew-hagemann:
Initial draft protofiles for provd:

Notes:
- Need to investigate landscape / wait on designs before we can tackle that protofile
- Only supporting ad out of the "identity providers"
- Left internet / networking out as this is covered by non-subiquity FE code currently
- EULA & a11y very much drafts and waiting on designs, so just best guess approximations for now 

Closes #197


provd/proto/accessibility.proto
```diff
@@ -0,0 +1,87 @@
+syntax = "proto3";
+
+package accessibility;
+
+service AccessibilityDirectoryService {
```
matthew-hagemann, line 5: I am assuming we would like these toggles to take immediate effect? So I have split them into individual requests and responses so we can update them via the screen and a user can test drive them.

provd/proto/keyboard.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package keyboard;
+
+service KeyboardService {
```
matthew-hagemann, line 5: My understanding is that the flutter frontend will already have access to all the keyboard types, so I have left the get out of this definition set.

provd/proto/theme.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package theme;
+
+service ThemeService {
```
matthew-hagemann, line 5: Again, split these into individual calls so that they can be called independently and the user can watch the effects take place.

provd/proto/accessibility.proto
```diff
@@ -0,0 +1,87 @@
...
+    rpc SetLargeText(SetLargeTextRequest) returns (SetLargeTextResponse) {}
+}
+
+message GetHighConstrastRequest {
+}
+
+message GetHighConstrastResponse {
+    bool success = 1;
+    string error = 2;
+    bool enabled = 3;
```
matthew-hagemann, line 22: In general, do we want to move forward with every call having a response with `success` and any error `messages`?

provd/proto/keyboard.proto
d-loose: It'd be nice to eventually remove those from the frontend though and be able to get a list of available keyboards via `GetKeyboard()` from the backend. Ideally we'd also have a keyboard detection mechanism like subiquity does, so I'd include `GetKeyboardStep()` in the API (of course we don't have to implement it immediately :D)

d-loose:
Looks great so far and is very much in line with the interfaces that we already have for the existing services :+1:.

I'd like to hear @tim-hm's and @didrocks' take on error handling, though. On the bootstrap side we essentially have
```dart
// subiquity_client/lib/src/client.dart

if (response.statusCode != 200) {
  throw SubiquityException(method, response.statusCode, str);
}
```
for unrecoverable errors and `GET meta/status` requests to subiquity which return one of:
```
STARTING_UP,
CLOUD_INIT_WAIT,
EARLY_COMMANDS,
NEEDS_CONFIRMATION,
WAITING,
RUNNING,
UU_RUNNING,
UU_CANCELLING,
DONE,
ERROR,
EXITED,
```

didrocks:
The error isn’t operated that way in GRPC. Contrary to JSON which doesn’t have rich status return code apart from the http statuses, GRPC have additional error so that you don’t transmit them inside the message itself. In this particular PR for instance, all errors are strings, and so you loose all the metadata and potential client side formatting from the error.

On GRPC, the underlying infrastructure is handling the transport layer error level. You still return an error directly to the GRPC call for any business-level error. However, if you want to differentiate from it on the client side, you can still catch the error status code. For instance, https://developers.google.com/actions-center/reference/grpc-api/status_codes returns the corresponding space for errors itself.

In Go, you should use https://pkg.go.dev/google.golang.org/grpc/status to convert and get from errors. You will find many articles on this https://jbrandhorst.com/post/grpc-errors/ and https://groups.google.com/g/golang-nuts/c/hiMjiDu37t8.

matthew-hagemann:
> The error isn’t operated that way in GRPC. Contrary to JSON which doesn’t have rich status return code apart from the http statuses, GRPC have additional error so that you don’t transmit them inside the message itself. In this particular PR for instance, all errors are strings, and so you loose all the metadata and potential client side formatting from the error.
> 
> On GRPC, the underlying infrastructure is handling the transport layer error level. You still return an error directly to the GRPC call for any business-level error. However, if you want to differentiate from it on the client side, you can still catch the error status code. For instance, https://developers.google.com/actions-center/reference/grpc-api/status_codes returns the corresponding space for errors itself.
> 
> In Go, you should use https://pkg.go.dev/google.golang.org/grpc/status to convert and get from errors. You will find many articles on this https://jbrandhorst.com/post/grpc-errors/ and https://groups.google.com/g/golang-nuts/c/hiMjiDu37t8.

Ah, so I can remove those success & message fields from all my responses, and in the case of where no data is being returned, I delete the message and substitute them with `google.protobuf.Empty` and we can just leverage that inbuilt status mechanism to get the same information? ie:

```
service AccessibilityService {
    // ... other RPCs ...

    rpc SetReducedMotion(SetReducedMotionRequest) returns (google.protobuf.Empty);
}

message SetReducedMotionRequest {
    bool enabled = 1;
}
```

didrocks:
Exactly! (in general, we do create our own Empty message to avoid the import and handling the dep if this is the only thing you import from google protobuf)

provd/proto/keyboard.proto
matthew-hagemann: I guess it falls in next to the networking logic. Ideally should be on the be, but pragmatically might need to stay on the fe for the time being :sweat_smile:

provd/proto/accessibility.proto
```diff
@@ -0,0 +1,48 @@
...
+
+service AccessibilityService {
+    rpc GetHighContract(Empty) returns (GetHighConstrastResponse) {}
+    rpc SetHighContract(SetHighConstrastRequest) returns (Empty) {}
+    rpc GetReducedMotion(Empty) returns (GetReducedMotionResponse) {}
+    rpc SetReducedMotion(SetReducedMotionRequest) returns (Empty) {}
+    rpc GetScreenReader(Empty) returns (GetScreenReaderResponse) {}
+    rpc SetScreenReader(SetScreenReaderRequest) returns (Empty) {}
+    rpc GetLargeText(Empty) returns (GetLargeTextResponse) {}
+    rpc SetLargeText(SetLargeTextRequest) returns (Empty) {}
```
tim-hm, line 13: In g-i-s today the accessibility service icon is shown (see screenshot below). I don't know what we need to cover in the UI, and will defer to @anasereijo, but I wanted to flag that this doesn't match up with it.

image

provd/proto/accessibility.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+
+message Empty {}
```
tim-hm, line 16: I seem to recall there being an Empty type available at `import "google/protobuf/Empty.proto"`.

Is this used or avoided @didrocks?

provd/proto/accessibility.proto
```diff
@@ -0,0 +1,48 @@
...
+
+package accessibility;
+
+service AccessibilityService {
+    rpc GetHighContract(Empty) returns (GetHighConstrastResponse) {}
+    rpc SetHighContract(SetHighConstrastRequest) returns (Empty) {}
+    rpc GetReducedMotion(Empty) returns (GetReducedMotionResponse) {}
+    rpc SetReducedMotion(SetReducedMotionRequest) returns (Empty) {}
+    rpc GetScreenReader(Empty) returns (GetScreenReaderResponse) {}
+    rpc SetScreenReader(SetScreenReaderRequest) returns (Empty) {}
```
tim-hm, line 11: `SetScreenReader` seems slightly off ... could just be me. We're not setting the screen reader per se, we're actually enabling / disabling it no?

`SetScreenReaderState` is slightly better imo, but you could also explore  `ActivateScreenReader`, `DeactivateScreenReader` and `GetScreenReaderState`.

I think a similar comment applies to the other methods.

provd/proto/accessibility.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+
+package accessibility;
+
+service AccessibilityService {
+    rpc GetHighContract(Empty) returns (GetHighConstrastResponse) {}
+    rpc SetHighContract(SetHighConstrastRequest) returns (Empty) {}
```
tim-hm, line 7: Typeo: Contract -> Contrast

provd/proto/ad.proto
```diff
@@ -0,0 +1,113 @@
...
+
+message CheckAdminNameRequest {
+    string admin_name = 1;
+}
+
+message CheckAdminNameResponse {
+    enum AdAdminNameValidation {
+    OK = 0;
+    EMPTY = 1;
+    INVALID_CHARS = 2;
```
tim-hm, line 48: indentation is off

provd/proto/ad.proto
```diff
@@ -0,0 +1,113 @@
...
+    rpc HasSupport(Empty) returns (HasSupportResponse);
+    rpc IsUsed(Empty) returns (IsUsedResponse);
+    rpc SetUsed(SetUsedRequest) returns (Empty);
+    rpc GetConnectionInfo(Empty) returns (GetConnectionInfoResponse);
+    rpc SetConnectionInfo(SetConnectionInfoRequest) returns (Empty);
+    rpc CheckDomainName(CheckDomainNameRequest) returns (CheckDomainNameResponse);
+    rpc CheckAdminName(CheckAdminNameRequest) returns (CheckAdminNameResponse);
+    rpc CheckPassword(CheckPasswordRequest) returns (CheckPasswordResponse);
+    rpc PingDomainController(PingDomainControllerRequest) returns (PingDomainControllerResponse);
+    rpc GetJoinResult(GetJoinResultRequest) returns (GetJoinResultResponse);
```
tim-hm, line 15: I suspect my questions/comments are because I don't know AD , but these method names could be improved. For example, "HasSupport", "Is Used" and "SetUsed" seem too generic to me and we can't assume the UI consuming this API will have deep AD knowledge. I have similar questions for many of the types here.

Lets jump on a call and discuss @matthew-hagemann?

provd/proto/eula.proto
```diff
@@ -0,0 +1,15 @@
...
+// Just a placeholder until we have designed the EULA
+service EULAService {
+    rpc AcceptEULA(AcceptEULARequest) returns (Empty) {}
+}
+
+message Empty {}
+
+message AcceptEULARequest {
+    string eulaId = 1;
+    string username = 2;
```
tim-hm, line 14: Is there a reason behind providing the username in the response and will there be more than one Eula on the machine which means we need to have an Id.

Perhaps a question for the OEM team; have you spoken with them?

Is convention here `AcceptEulaRequest` or `AccpetEULAReqest?`

provd/proto/identity.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+
+package identity;
+
+service IdentityService {
+  rpc GetIdentity(Empty) returns (GetIdentityResponse);
+  rpc SetIdentity(SetIdentityRequest) returns (Empty);
+  rpc ValidateUsername(ValidateUsernameRequest) returns (ValidateUsernameResponse);
```
tim-hm, line 8: Whats the purpose of this call? Ie we validate username but not password / hostname / realname?

provd/proto/identity.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+message Empty {}
+
+message Identity {
+  string realName = 1;
```
tim-hm, line 14: Something about the capitalised `N` 🤔

provd/proto/identity.proto
```diff
@@ -0,0 +1,35 @@
...
+  string hostname = 2;
+  string username = 3;
+  string password = 4;
+  bool autoLogin = 5;
```
tim-hm, line 18: this value doesn't really seem right in an 'Identity' object

provd/proto/identity.proto
```diff
@@ -0,0 +1,35 @@
...
+message SetIdentityRequest {
+  Identity identity = 1;
+}
+
+message ValidateUsernameRequest {
+  string username = 1;
+}
+
+message ValidateUsernameResponse {
+  bool validation = 1;
```
tim-hm, line 34: 'valdiation' could be ambiguous as a boolean. For example, does `validation: false` mean validation didn't run or that the input wasn't valid?

Maybe simply: `bool valid = 1;`?

provd/proto/keyboard.proto
```diff
@@ -0,0 +1,25 @@
...
+    rpc SetKeyboard(SetKeyboardRequest) returns (Empty) {}
+    rpc SetInputSource(SetInputSourceRequest) returns (Empty) {}
+}
+
+message Empty {}
+
+message KeyboardSettings {
+    string layout = 1;
+    string varient = 2;
```
tim-hm, line 14: typeo: varient -> variant

provd/proto/keyboard.proto
```diff
@@ -0,0 +1,25 @@
...
+}
+
+message SetKeyboardRequest {
+    KeyboardSettings settings = 1;
+}
+
+message SetInputSourceRequest {
+  KeyboardSettings settings = 1;
+  string user = 2;
```
tim-hm, line 23: Is"user" necessary? When its running before the first user is created there is no user yet ... and if its the 2nd user that's created surely we can infer which user this relates to?

provd/proto/privacy.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package privacy;
+
+service PrivacyService {
+    rpc SetPrivacySettings (SetPrivacySettingsRequest) returns (Emtpy) {}
```
tim-hm, line 6: The approach here is different to the approach in a11y. Here its an objective for all of the options, but in the a11y section its option by option.

Iim not sure if one approach is better than the other, but we should be consistent in either case.

provd/proto/privacy.proto
```diff
@@ -0,0 +1,15 @@
...
+}
+
+message Empty {}
+
+message SetPrivacySettingsRequest {
+    bool locationEnabled = 1;
+    bool reportingEnabled = 2;
```
tim-hm, line 13: I think we can do better with both names here. For example, is 'reportingEnabled' about ongoing bug reporting, one off installation metrics, or something else entirely?

provd/proto/pro_enrolment.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package pro_enrollment;
+
+service ProEnrollmentService {
+    rpc GenerateEnrolmentCode(GenerateEnrolmentCodeRequest) returns (GenerateEnrolmentCodeResponse);
```
tim-hm, line 6: You've spelt enrolment in two ways here :)

provd/proto/pro_enrolment.proto
```diff
@@ -0,0 +1,23 @@
...
+    rpc SetTokenManually(SetTokenManuallyRequest) returns (Empty);
```
tim-hm, line 7: Is there a way to automatically set the token?

provd/proto/theme.proto
```diff
@@ -0,0 +1,33 @@
...
+    rpc SetBrightness(SetBrightnessRequest) returns (Empty);
+    rpc GetBrightness(Empty) returns (GetBrightnessResponse);
```
tim-hm, line 7: I think the light/dark theme is call 'variant'? Brightness is too similar to screen brightness imo.

provd/proto/theme.proto
```diff
@@ -0,0 +1,33 @@
...
+message SetBrightnessRequest {
+    Brightness brightness = 1;
+}
+
+message GetBrightnessResponse {
+    Brightness brightness = 1;
+}
+
+message SetAccentRequest {
+    string accent = 1;
```
tim-hm, line 28: There's a finite number of accent colours so it might be worth putting those into an enum?

tim-hm (changes requested):
A bunch of comments in here for you @matthew-hagemann. Have a read and maybe we can jump on a call to walk through?

provd/proto/accessibility.proto
anasereijo: Good points, @tim-hm! I am yet to define what we should display here. On Elio's [design](https://www.figma.com/file/xkI4paNpSjKhAq9cHhrRXx/24.04-Desktop-Provisioning-UX?type=design&node-id=539%3A4975&mode=design&t=UQY1oOqHzVhgjO5l-1) less a11y settings are displayed. However, after looking at what you just shared from the a11y dropdown, I think we can include all those options! We can even then group them to make it look nicer and easier to scan. @matthew-hagemann how costly it is for you to add all options even if there's a chance you will not include them all in the final design?

provd/proto/accessibility.proto
didrocks: The real answer is "it depends" :)

More generally, I don’t know on any best practice here, but the rule of thought here is following the Go proverb "a little copying is better than a little dependency" (https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s). If we are only importing from google protobuf "Empty", let’s try to avoid this and copy the empty message type (this is what we do in other projects). This is in the end easier to setup, less risk of missing the dep and so on…

Once we are importing more proto files from this repo, then, we can move to the google’s version, which is binary compatible anyway :)

provd/proto/accessibility.proto
matthew-hagemann: Fairly easy! I will add more and we can pick and choose what to implement when the time comes :slightly_smiling_face:

provd/proto/ad.proto
matthew-hagemann: I'll do some digging on what all these settings actually do to see if we can improve the naming past what was used in the abstract classes :+1:

provd/proto/eula.proto
matthew-hagemann: Getting final clarification from the OEM team, but for now it doesn't look like we need to actually pass anything in

provd/proto/keyboard.proto
matthew-hagemann: I see in GIS there seems to be a check for when we are creating a new user, and if we are it sets the settings against that users GSettings. Otherwise it will attempt a system wide setting of the keyboard. 

Looking at the dart code, it looks like we grab the user from environment variables when setting it at a user level, so it should be safe to remove this and move it to the BE :+1:

provd/proto/privacy.proto
matthew-hagemann: Looking at the flow screenshots / dart pages, I see that although the model includes the reporting toggle, it is not actually implemented in page yet.

d-loose (approved):

PR #228, opened 2023-11-22
feat(ci): run tests only on main and archive/main

d-loose:
We don't need to run the Flutter and Go tests on the snap branch - instead just run the snap workflow there.

I'd open a PR that merges `main` into `bootstrap-snap` afterwards.

P.S. I've also removed the obsolete `canary-*` and `ubuntu/**` branches

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,12 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


spydon (approved):

d-loose:
Oh, you got write-access for real now so your reviews are finally "valid" :D

PR #232, opened 2023-11-23
ci(codecov): Add change failour thresholds

### LINKED ISSUE #209: Workflow fails on codecov decreases
Codecov deceasing by a fraction of a percentage fails the workflow. We should remove this condition so that we don't get used to seeing a failing workflow, and leave it instead to reviewer / PR opener to reason with manually.

matthew-hagemann:
Added a 5% threshold to the codecov check
Added a minimum target of 80% to keep us from dropping too far

Closes #209


d-loose (approved):
Thanks, LGTM :+1:

PR #237, opened 2023-11-27
refactor: Split up widgets to files

spydon:
Currently it is quite hard to navigate since there was a multitude of widgets and layouts within the same files, this splits them up and puts them into directories. It also creates barrel files for them to not make the imports harder.


d-loose (approved):
LGTM :+1:

Issue #205, opened 2023-11-17
Draft white-labelling support for the UIs


d-loose:
To meet the requirements from OEM, other Ubuntu flavors, Core Desktop, etc., we need to support a certain level of UI customization. This should potentially include customizing:
- accent/primary color
- additional colors(?)
- font(?)
- window title
- image assets
- installation slide show(?)

PR #239, opened 2023-11-28
feat: add theme variant service

d-loose:
Adds a `ThemeVariantService` to ubuntu-provision that loads some configuration options from the `ConfigService` and creates a `ThemeVariant`, which contains customized yaru themes and a window title.
Example configuration:
```yaml
bootstrap:
  accent-color: '#0000ff'
  elevated-button-color: '#ff0000'
  elevated-button-text-color: '#00ff00'
  window-title: 'My custom window title'
```
(the color parameters are already supported by yaru.dart without any further customization needed).

I'm sure we'll want to adjust details like config file location, scope, etc later on.

Ref #205
UDENG-1762

### DIFF
--- a/packages/ubuntu_provision/lib/src/services/theme_variant_service.dart
+++ b/packages/ubuntu_provision/lib/src/services/theme_variant_service.dart
@@ -1,10 +1,29 @@
 import 'package:flutter/material.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:yaml/yaml.dart';
 import 'package:yaru/yaru.dart';
 
 import '../../services.dart';
 
+part 'theme_variant_service.freezed.dart';
+part 'theme_variant_service.g.dart';
+
 final _log = Logger('theme_variant');
+
+@freezed
+class ThemeConfig with _$ThemeConfig {
+  // ignore: invalid_annotation_target
+  @JsonSerializable(fieldRename: FieldRename.kebab, explicitToJson: true)
+  const factory ThemeConfig({
+    String? accentColor,
+    String? elevatedButtonColor,
+    String? elevatedButtonTextColor,
+    String? windowTitle,
+  }) = _ThemeConfig;
+  factory ThemeConfig.fromJson(Map<String, dynamic> json) =>
+      _$ThemeConfigFromJson(json);
+}
 
 class ThemeVariant {
   const ThemeVariant({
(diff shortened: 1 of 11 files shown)


packages/ubuntu_provision/lib/src/services/theme_variant_service.dart
```diff
@@ -0,0 +1,77 @@
+import 'package:flutter/material.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:yaru/yaru.dart';
+
+import '../../services.dart';
+
+/// @internal
```
spydon, line 7:
```suggestion
@internal
```
`meta` will have to be imported too.

Will this be used outside of this file? Otherwise it could just be made private.

d-loose: I just wanted to be consistent with all the other files that follow the exact same pattern. Is it worth changing this across the entire project?

spydon: What does a dartdoc with @internal in it even mean? Surely that doesn't affect the analyzer? :thinking:

d-loose: I have no idea, it's probably unnecessary and doesn't do anything, since the `log` symbols need to be explicitly hidden from the barrel exports

spydon: Let's make them private where possible and `@internal` at the other places then, I can do that in a separate PR. :)

tim-hm:
Great start @d-loose 🙏 🤩  I think we should start with a simpler approach and evolve if/when required; after all YAGNI and its harder to retract the feature once its live! I think we should require/assume that if a `whitelabel.yaml` was provided then it will be available for every stage -- ie we copy it from bootstrap to the target so its automatically picked up during 4/5/6.

A few suggestions on the schema:

1. The styling override could be against a global theme key and independent of the stage we're on:

```
theme:
  accent-color: '#0000ff'
  elevated-button-color: '#ff0000'
  elevated-button-text-color: '#00ff00'
```

2. For overriding the window title and image we can then work by stage / screen:

```
screens:
   name-of-step:
     image: <path-to-image>
     title: <custom title>
```

3. Finally, for turning on/off individual screens we simply provide false aginst the step id/name:

```
screens:
   name-of-step: false
```

WDYT?

d-loose:
> I think we should require/assume that if a `whitelabel.yaml` was provided then it will be available for every stage -- ie we copy it from bootstrap to the target so its automatically picked up during 4/5/6.

Yes, that makes sense. Will we need more than one configuration file for the UI in the end? We're currently assuming there's a single `ubuntu-provision.yaml`, which we could just rename to `whitelabel.yaml` if that's all we need. All the other configuration files (autoinstall, dry-run, etc) are read by subiquity.


> A few suggestions on the schema:
> ...
> WDYT?

Looks good!

tim-hm:
> Will we need more than one configuration file for the UI in the end? 

I think we should be working towards a single whitelabel.yaml file for the moment.

d-loose:
We're currently using `bootstrap` and `init` scopes in the config file - I'd restructure this and adopt the suggested schema in a separate PR.

d-loose:
Since I did some refactoring, it seems to make more sense to include the restructuring in this PR.
On the surface, not much has changed:
```yaml
theme:
  accent-color: '#0000ff'
  elevated-button-color: '#ff0000'
  elevated-button-text-color: '#00ff00'
  window-title: 'My custom window title'
```
can be used to configure the theme for both ubuntu-bootstrap and ubuntu-init now (I've removed the separate config scopes).

I've added `freezed` and `json_serialize` to ubuntu_provision for a more structured config parsing approach. It might make sense to move the `ThemeConfig` class out of the service at a later point.

PR #244, opened 2023-12-01
feat: add PageConfigService

d-loose:
Follow-up to #239.
Adds a service that parses the 'pages' section of the configuration file following this schema:

```yaml
pages:
  locale: false
  keyboard:
    image: '/path/to/image'
    title: 'My custom window title'
```

This PR only replaces the previous 'pages' configuration that allowed passing a list of pages that should be included in the wizard. As before, by default all available pages are shown by default. What's changed is that the logic is now reversed: you need to exclude a page explicitly in the config file now. If a page is excluded but required by subiquity in autoinstall mode, a warning will be shown and only the pages by subiquity will be included. (We could also exit the wizard with an error, if that's preferred).

As a next step, I plan to add a provider that holds the parsed page config so that individual pages have a way to access the customized image and window title.

P.S. I'll add proper integration tests once we settle on a config schema!

Fix: #205
UDENG-1762

### DIFF
--- a/packages/ubuntu_bootstrap/lib/installer.dart
+++ b/packages/ubuntu_bootstrap/lib/installer.dart
@@ -28,7 +28,6 @@ export 'slides.dart';
 
 Future<void> runInstallerApp(
   List<String> args, {
-  List<String>? pages,
   List<WidgetBuilder>? slides,
   UbuntuFlavor? flavor,
   ThemeData? theme,
@@ -53,8 +52,6 @@ Future<void> runInstallerApp(
         defaultsTo: 'examples/sources/desktop.yaml',
         help: 'Path of the source catalog (dry-run only)');
     parser.addFlag('welcome', aliases: ['try-or-install'], hide: true);
-    parser.addOption('pages',
-        valueHelp: 'path', help: 'A comma-separated list of pages', hide: true);
   })!;
   final liveRun = options['dry-run'] != true;
   final exe = p.basename(Platform.resolvedExecutable);
@@ -84,8 +81,7 @@ Future<void> runInstallerApp(
   tryRegisterServiceFactory<GSettings, String>((schema) => GSettings(schema));
   tryRegisterService<InstallerService>(() => InstallerService(
       getService<SubiquityClient>(),
-      config: tryGetService<ConfigService>(),
-      pages: (options['pages'] as String?)?.split(',') ?? pages));
+      pageConfig: tryGetService<PageConfigService>()));
   tryRegisterService(JournalService.new);
   tryRegisterService<KeyboardService>(
       () => SubiquityKeyboardService(getService<SubiquityClient>()));
@@ -93,6 +89,8 @@ Future<void> runInstallerApp(
       () => SubiquityLocaleService(getService<SubiquityClient>()));
   tryRegisterService<NetworkService>(
       () => SubiquityNetworkService(getService<SubiquityClient>()));
+  tryRegisterService(
+      () => PageConfigService(config: tryGetService<ConfigService>()));
   tryRegisterService(() => PostInstallService('/tmp/$baseName.conf'));
   tryRegisterService(PowerService.new);
   tryRegisterService(ProductService.new);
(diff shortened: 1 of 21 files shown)


packages/ubuntu_bootstrap/lib/installer.dart
```diff
@@ -53,8 +52,6 @@ Future<void> runInstallerApp(
         defaultsTo: 'examples/sources/desktop.yaml',
         help: 'Path of the source catalog (dry-run only)');
     parser.addFlag('welcome', aliases: ['try-or-install'], hide: true);
-    parser.addOption('pages',
-        valueHelp: 'path', help: 'A comma-separated list of pages', hide: true);
```
spydon, line 57: Were we going to drop command line support completely?

packages/ubuntu_bootstrap/lib/services/installer_service.dart
```diff
@@ -33,16 +36,28 @@ class InstallerService {
 
   Future<void> load() async {
     await monitorStatus().firstWhere((s) => s?.isLoading == false);
-    _pages ??=
-        await _client.getInteractiveSections() ?? await _config?.getPages();
+    _subiquityPages = (await _client.getInteractiveSections())?.toSet();
+    _excludedPages = _pageConfig?.excludedPages;
+
+    if ((_subiquityPages?.isNotEmpty ?? false) &&
+        (_excludedPages?.isNotEmpty ?? false)) {
```
spydon, line 43: Since they are sets you could just do the intersection of the sets and then loops over that

packages/ubuntu_bootstrap/lib/services/installer_service.dart
```diff
@@ -33,16 +36,28 @@ class InstallerService {
...
+      final requiredExcludedPages =
+          _excludedPages!.where(_subiquityPages!.contains);
+      _log.info(
+          '$requiredExcludedPages are required by subiquity and cannot be excluded!');
+      _excludedPages!.removeWhere(requiredExcludedPages.contains);
```
spydon, line 48: And then you can do removeAll with the intersection set here

spydon (commented):
A few minor comments, but nothing that affects the outcome

packages/ubuntu_bootstrap/lib/installer.dart
d-loose: I don't think it makes sense to have a command line option that allows to *include* specific pages, while we're using the config entries to *exclude* pages.
I'd come back to command line options at a later point - maybe we'll want to have both `--exclude-pages` and `--include-pages` or something.

packages/ubuntu_bootstrap/lib/services/installer_service.dart
d-loose: Thanks, good point! Didn't think about what further improvements I could make when I changed those types from lists to sets

packages/ubuntu_bootstrap/lib/services/installer_service.dart
```diff
@@ -33,16 +36,28 @@ class InstallerService {
...
+    _excludedPages = _pageConfig?.excludedPages;
+
+    if ((_subiquityPages?.isNotEmpty ?? false) &&
+        (_excludedPages?.isNotEmpty ?? false)) {
+      final requiredExcludedPages =
+          _excludedPages!.intersection(_subiquityPages!);
+      _log.info(
+          '$requiredExcludedPages are required by subiquity and cannot be excluded!');
+      _excludedPages!.removeAll(requiredExcludedPages);
+    }
```
spydon, line 49:
```suggestion
    final requiredExcludedPages =
          _excludedPages?.intersection(_subiquityPages ?? {});
    if (requiredExcludedPages.isNotEmpty) {
      _log.info(
          '$requiredExcludedPages are required by subiquity and cannot be excluded!');
      _excludedPages!.removeAll(requiredExcludedPages);
    }
```
Super nitpicking :sweat_smile:
Just improves readability a bit.

spydon (approved):

PR #245, opened 2023-12-04
chore: Bump dependencies

spydon:


packages/ubuntu_init/test/init_wizard_test.dart
```diff
@@ -89,6 +89,7 @@ void main() {
     await tester.pumpAndSettle();
     expect(find.byType(NetworkPage), findsOneWidget);
     verify(networkModel.init()).called(1);
+    await tester.pumpAndSettle();
```
d-loose, line 92: Hm, weird that `pumpAndSettle()` gets essentially called twice in a row here - is there some underlying problem in the `NetworkPage`? This doesn't seem to be necessary for any other page in the init wizard test.

spydon: I think it is because it emulates some waiting time for when the available networks should pop up or something like that

d-loose (approved):
LGTM :+1:

Issue #252, opened 2023-12-05
Setup temporary pipeline for testing


d-loose:
In order to test the entire new provisioning flow, we need to modify the target system according to our needs (probably using subiquity's post-install hooks):

- Override default user creation in cloud.cfg
- Replace gnome-initial-setup with ubuntu-init

PR #257, opened 2023-12-06
feat: add script to install ubuntu init

d-loose:
This postinstall script creates a script on the target system that fetches the ubuntu-desktop-init snap from a [test repo](https://github.com/d-loose/provision-testing) and installs it. Its execution is triggered by a systemd unit that runs after the network is initialized and snapd has finished seeding and before the display manager starts. Lastly, it overrides the `Exec=` line in `gnome-initial-setup.desktop`, making it point directly to the `ubuntu_init` binary from the snap (i.e. it runs outside any snap environment).

Ref: #252
UDENG-1913

### DIFF
--- a/snap/local/postinst.d/99_install_ubuntu_init
+++ b/snap/local/postinst.d/99_install_ubuntu_init
@@ -1,0 +1,33 @@
+#!/bin/bash
+TARGET=/target
+
+set -eux
+
+(
+cat << 'EOF'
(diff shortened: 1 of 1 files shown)


spydon (approved):
Lgtm! Tested and it takes 25s after boot for it to pop up on my VM, and it can't get past the language screen but I guess that is unrelated.

PR #258, opened 2023-12-07
feat(init): catch/log xdg service exceptions and reorder pages

d-loose:
This makes the existing xdg services a bit more robust so that we can use them to test the new flow instead of replacing them with fake services.
We can successively replace those services with calls to the backend in the future.

I've also reordered the pages to be more in-line with the new flow. The store page isn't supposed to be part of the first-boot-initialization flow - I'll move it to a separate 'desktop welcome' wizard (to be run when the new user first logs in) once we implement it.

Ref: #252
UDENG-1913

### DIFF
--- a/packages/ubuntu_init/test/init_wizard_test.dart
+++ b/packages/ubuntu_init/test/init_wizard_test.dart
@@ -89,7 +89,6 @@ void main() {
     await tester.pumpAndSettle();
     expect(find.byType(NetworkPage), findsOneWidget);
     verify(networkModel.init()).called(1);
-    await tester.pumpAndSettle();
 
     await tester.tapNext();
     await tester.pumpAndSettle();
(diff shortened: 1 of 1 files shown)


d-loose:
The init wizard test is acting up again. I think the underlying issue could be the asset pre-caching for the `TimezonePage` - I'll have a look on Monday.

spydon (commented):
There isn't anything depending on the Future errors I guess currently? Since you're not re-throwing?

d-loose:
Well at least the UI doesn't care whether what the page is supposed to accomplish was successful or not. I'm deliberately not rethrowing any exceptions here so that we test the existing pages and models without getting stuck in the UI.

Of course we need to focus on meaningful error handling once we start replacing those services with new ones that communicate with the backend.

spydon (approved):
After clarifcation; Lgtm!

matthew-hagemann:
@d-loose definitely! I'll look into it now and push the update for you to re-base off of if you are happy to wait a few?

PR #259, opened 2023-12-08
feat: Provd user client and structure of provd_client package

spydon:
This PR introduces the basic structure of the  provd_client package.
In this PR we assume that there will be multiple client classes for different parts of connections with provd, for example the `ProvdUserClient`, which is the only one created so far.


packages/provd_client/build/test_cache/build/c075001b96339384a97db4862b8ab8db.cache.dill.track.dill
d-loose, line 1: should probably be .gitignored

d-loose: err, I mean the entire `build` directory

spydon: Good catch, I'll create a global gitignore so that each package doesn't have to put up their own rules for things like this.

Issue #248, opened 2023-12-05
User service impl


matthew-hagemann:
To unblock the first page of work for the FE team, we need to implement the User service in the provd go BE

PR #260, opened 2023-12-11
User service impl

### LINKED ISSUE #261: Go dbus mocking framework
We need to import / adapt the dbus mocking framework from one of the other go projects to correctly mock dbus for testing purposes

matthew-hagemann:
Closes #261, #248


provd/internal/services/user/user.go
```diff
@@ -0,0 +1,293 @@
...
+	// Set the hostname
+	err = s.hostname.Call("org.freedesktop.hostname1.SetStaticHostname", 0, hostname, false).Err
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to set hostname: %s", err)
+	}
+
+	return &proto.Empty{}, nil
+}
+
+func (s *Service) ValidateUsername(ctx context.Context, req *proto.ValidateUsernameRequest) (*proto.ValidateUsernameResponse, error) {
```
d-loose, line 173: This method only checks if the given username already exists, but there are a few more cases we need to consider, see [implementation in subiquity](https://github.com/canonical/subiquity/blob/e3a0f6d215fdd19deaa854bdec0298b4a5b57ee3/subiquity/server/controllers/identity.py#L122-L135).

I'm wondering what's the best way to return the validation result - a boolean plus a specific error in case of failure, or an enum like in subiquity?

provd/internal/services/user/user.go
```diff
@@ -0,0 +1,293 @@
...
+		// Unknown error
+		return nil, status.Errorf(codes.Internal, "unknown error: %v", err)
+	}
+
+	// User found
+	return &proto.ValidateUsernameResponse{Valid: false}, nil
+
+}
+
+func (s *Service) GetUser(ctx context.Context, req *proto.GetUserRequest) (*proto.GetUserResponse, error) {
```
d-loose, line 204: As far as I understand we don't really need this method (and could even remove it from the protos).

Subiquity's `GetIdentity()` doesn't take any arguments and just returns an identity if it has already been configured before.

provd/internal/services/user/user.go
```diff
@@ -0,0 +1,293 @@
...
+	hasher := sha512.New()
+	hasher.Write(salt)
+	hasher.Write([]byte(password))
+	hashedBytes := hasher.Sum(nil)
+	hashedPassword := base64.RawStdEncoding.EncodeToString(hashedBytes)
+
+	// Format the hash in the SHA-512 crypt format
+	saltStr := base64.RawStdEncoding.EncodeToString(salt)
+	return fmt.Sprintf("$6$%s$%s", saltStr, hashedPassword), nil
+}
```
d-loose, line 103: This seems to generate invalid password hashes. I think the issue is that the random 16-byte `salt` is base64 encoded, resulting in a `saltStr` longer than 16 bytes. As I understand it the length of the salt string (between the '$' symbols has to be between 8 and 16 (inclusive) characters long. So it's not 16 random bytes but 16 random characters (from [this set](https://github.com/maximeh/buildroot/blob/master/package/mkpasswd/mkpasswd.c#L74-L75)).
Sample outputs of `mkpasswd -m sha512crypt` with password 'foo':
```
$6$Cl3G4BBXctuRu98R$m0QWgf9doBLkBGryt8IAGpKgJDz0Jpls1R4fRAGdMnT7MT8a22ufkTYofct68FynDCOeYAr4UQiONYniKLnYa/
$6$tS0n2eZeF0IubeMY$n/X.zz7QHntZDjZqmYt86Oa/61w1GY3cKNX6R.G4MU.vn7DRv3bHiNK0CeZlGDE2H9AH4LnRTHmecPJyWpzvy/
$6$sXz.U7C4ktDZrytw$FfPdFu9pPQoWQ9oVXABtRb9ZXwPTiOs0TpKlRGbNx0kSqrpRAuIIllpFuHJu2ItyC.mGobuAnwo.3bnP/NqXO0
```

provd/internal/services/user/user.go
matthew-hagemann: Cool, I'll clean it up when I do my final code pass :slightly_smiling_face:

provd/internal/services/user/user.go
matthew-hagemann: I've updated the impl to check those additional restrictions :+1: 

Now that it accounts for more failure cases, using an emun to represent the outcome makes sense to me :slightly_smiling_face: we are going to use the same pattern for ad validation aswell.

provd/proto/user.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+message Empty {}
+
+message GetUserRequest {
+  string user_id = 1;
+}
+
+message User {
+  string realName = 1;
```
spydon, line 20: All field names should be snake case I think?
https://protobuf.dev/programming-guides/style/#message-field-names

provd/proto/user.proto
```diff
@@ -0,0 +1,51 @@
...
+  string hostname = 2;
+  string username = 3;
+  string password = 4;
+  bool autoLogin = 5;
```
spydon, line 24: And this

provd/proto/user.proto
```diff
@@ -0,0 +1,51 @@
...
+  User user = 1;
+  bool is_admin = 2;
+}
+
+message ValidateUsernameRequest {
+  string username = 1;
+}
+
+message ValidateUsernameResponse {
+  UsernameValidation usernameValidation = 1;
```
spydon, line 41: And this

provd/proto/user.proto
matthew-hagemann: Ah yes! It should all be snake :smile:

matthew-hagemann: Should be, Go internally changes it all to PascalCase so I think the generator maps it to whatever is idiomatic for that language :slightly_smiling_face:

matthew-hagemann:
@spydon @d-loose I am going through this again with Didier on Friday in a pair programming session, but changes should be minimal.

provd/protos/user.proto
```diff
@@ -38,5 +29,14 @@ message ValidateUsernameRequest {
...
+  UsernameValidation username_validation = 1;
 }
+
+enum UsernameValidation {
+    OK = 0;
+    ALREADY_IN_USE = 1;
+    SYSTEM_RESERVED = 2;
+    INVALID_CHARS = 3;
+    TOO_LONG = 4;
+    EMPTY = 5;
```
d-loose, line 41: Nitpick: the indentation level is different here than in the other blocks

matthew-hagemann: Fair! I see the style guide says 2 spaces, which is what I've done everywhere else, so I'll just align this :+1:

d-loose (approved):
Thanks, LGTM :+1:

PR #266, opened 2023-12-11
chore: Fix analyzer issues reported by ubuntu_lints 0.2.0

spydon:


packages/subiquity_test/lib/src/utils.dart
```diff
@@ -1,6 +1,6 @@
 import 'dart:io';
 
-import 'package:subiquity_test/subiquity_test.dart';
+import '../subiquity_test.dart';
```
d-loose, line 3: How do you feel about using import paths relative to the package root, like we do in app-center (i.e. `import '/subiquity_test.dart'`)? It makes moving files around within a package much cleaner - does it have any disadvantages?

spydon: Usually people choose between package imports or relative (never seen anyone using absolutes), we could use package imports for local ones too to accommodate for that. That's what we do in the flame repo.

There is a linter rule for that too.

d-loose: Ah, didn't know that was possible, I like that!

packages/subiquity_client/lib/src/types.dart
```diff
@@ -1,3 +1,6 @@
+// TODO(Lukas): Rename enums to dart style
+// ignore_for_file: constant_identifier_names
+
```
d-loose, line 3: Could we also ignore `always_put_required_named_parameters_first` in this file and move the `sizingPolicy` back? Otherwise we'd need update generator.py accordingly.

packages/subiquity_test/lib/src/utils.dart
spydon: https://github.com/canonical/ubuntu-flutter-plugins/pull/367

packages/subiquity_client/lib/src/types.dart
spydon: Does the generator not taking into consideration that these are named arguments and not positional ones?

packages/subiquity_test/lib/src/utils.dart
d-loose: Hm sorry, I think I misunderstood your comment I thought there was a linter rule for those absolute path imports now (`prefer_relative_imports` doesn't complain about those, but the auto-fix options only offer package imports or relative ones). I'm not sure if it's worth converting all imports to package imports, since we've been using relative ones mostly. What do you prefer?

packages/subiquity_client/lib/src/types.dart
d-loose: The generator is really simple: it traverses subiquity's types.py and generates Dart code with all parameters in the same order. It marks a parameter as `required` if the corresponding default value is `None` or something that can't be converted into a primitive Dart type.

packages/subiquity_test/lib/src/utils.dart
d-loose: Ok, I see. If you've had good experience with package imports I'd also be fine using those. (Let's hope we won't have to rename packages too often again :D)

packages/subiquity_client/lib/src/types.dart
spydon: Ah, makes sense, I'll revert that change and add an ignore!

d-loose: Thank you!

packages/ubuntu_provision_test/lib/src/wizard_tester.dart
```diff
@@ -36,7 +39,7 @@ extension UbuntuWizardTester on WidgetTester {
 
   Future<void> jumpToPage(String route) async {
     final context = element(find.byType(WizardPage));
-    Wizard.of(context).jump(route);
+    await Wizard.of(context).jump(route);
```
d-loose, line 42:
```suggestion
    unawaited(Wizard.of(context).jump(route));
```
should fix the screenshot tests (needs a `dart:async` import)

d-loose (approved):

Issue #255, opened 2023-12-05
Adopt new provisioning flow for ubuntu-init


d-loose:
We should check if the existing pages in the init wizard need to be re-ordered and add placeholder pages for the missing pages. It's probably a good idea to focus on the 'home' flow first (user creation -> pro enrollment -> location services -> timezone selection -> telemetry -> theme selection). The 'work' flow involves a more complicated user creation / authentication step - I'd do that in a second step.

PR #273, opened 2023-12-12
feat: add placeholder page for ubuntu pro and remove store page

d-loose:
Adds a blank page for Pro enrollment and removes the 'store' page, which is now obsolete and won't be needed in either bootstrap or init.

Ref #255
UDENG-1915


spydon (approved):
Lgtm!

PR #278, opened 2023-12-13
fix: catch all errors in init services

d-loose:
The debug error catching introduced in #258 got messed up in #272. This reverts those changes so that we can run the init wizard using those legacy services without crashing it.
I've added todo notes to the services that we'll need to deprecate and replace with calls to the provd daemon.

### DIFF
--- a/packages/ubuntu_init/lib/src/services/privacy_service.dart
+++ b/packages/ubuntu_init/lib/src/services/privacy_service.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: avoid_catches_without_on_clauses
+
 import 'package:dbus/dbus.dart';
 import 'package:gsettings/gsettings.dart';
 import 'package:meta/meta.dart';
@@ -32,7 +34,7 @@ class GnomePrivacyService implements PrivacyService {
   Future<bool> isLocationEnabled() async {
     try {
       return await _locationSettings.get('enabled').then((v) => v.asBoolean());
-    } on Exception catch (e) {
+    } catch (e) {
       _log.error('Error getting location settings: $e');
       return false;
     }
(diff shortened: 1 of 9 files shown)


packages/ubuntu_init/lib/src/services/privacy_service.dart
```diff
@@ -1,3 +1,6 @@
+// ignore_for_file: avoid_catches_without_on_clauses
```
spydon, line 1: Instead of doing ignore we can disable it in the local analysis file, or are there places where we use catches properly too?

d-loose: Most of the try-catch blocks use a proper `on` clause, so I'd keep the rule active globally and deal with the cases where it's messing.

I'd remove the recently added `on Exception` clauses in the [product service](https://github.com/canonical/ubuntu-desktop-provision/blob/675f89cf5a24c72ba1736d0f989bd7e866b7a9b1/packages/ubuntu_provision/lib/src/services/product_service.dart#L85-L110) and in the refresh service ([here](https://github.com/canonical/ubuntu-desktop-provision/blob/675f89cf5a24c72ba1736d0f989bd7e866b7a9b1/packages/ubuntu_bootstrap/lib/services/refresh_service.dart#L58) and [here](https://github.com/canonical/ubuntu-desktop-provision/blob/675f89cf5a24c72ba1736d0f989bd7e866b7a9b1/packages/ubuntu_bootstrap/lib/services/refresh_service.dart#L72)) and add TODO notes there as well. I don't want to only catch a smaller subset of errors without checking that it doesn't break anything first, but we should definitely clean those up.

spydon (approved):

PR #282, opened 2023-12-18
deps: Remove root ubuntu_lints

spydon:
The `ubuntu_lints` dependency should live in all packages, not in the root.


d-loose (approved):
Since the pubspec.lock is checked into the repo now, it needs to be updated as well. Besides that, LGTM :+1:

PR #285, opened 2023-12-19
feat: Pre-cache all images from the PageConfigProvider

spydon:
This PR adds a image cache that loads all images specific by the white labeling config, the loading page is shown meanwhile.
It also speeds up some loading by starting to run multiple futures at once instead of running the sequentially.


packages/ubuntu_bootstrap/lib/installer.dart
```diff
@@ -78,6 +78,8 @@ Future<void> runInstallerApp(
   );
   tryRegisterService<DesktopService>(GnomeService.new);
   tryRegisterServiceFactory<GSettings, String>(GSettings.new);
+  tryRegisterService(
+      () => PageConfigService(config: tryGetService<ConfigService>()));
```
d-loose, line 82: Could we keep those services in alphabetic order?

packages/ubuntu_bootstrap/test/test_utils.dart
```diff
@@ -1,11 +1,15 @@
...
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mockito/annotations.dart';
+import 'package:mockito/mockito.dart';
 import 'package:ubuntu_bootstrap/l10n.dart';
 import 'package:ubuntu_bootstrap/services.dart';
+import 'package:ubuntu_provision/ubuntu_provision.dart';
 import 'package:ubuntu_wizard/ubuntu_wizard.dart';
 import 'package:yaru/yaru.dart';
 
+import 'confirm/test_confirm.dart';
```
d-loose, line 11: We should probably import `'../../ubuntu_provision/test/test_utils.mocks.dart'` here

spydon: Ah yes, in the long run we should probably create an internal package in the monorepo with test helpers, so that we don't have to import files like this.

packages/ubuntu_provision/lib/src/services/page_config_service.freezed.dart
```diff
@@ -138,7 +138,7 @@ class _$PageConfigEntryImpl implements _PageConfigEntry {
   }
 
   @override
-  bool operator ==(dynamic other) {
+  bool operator ==(Object other) {
```
d-loose, line 141: Are those unintended changes caused by the linter? Freezed seems to generate those lines with a `dynamic` type.

spydon: The new version generates `Object` instead of `dynamic`, it's strange that it isn't `Object?` though...

packages/ubuntu_bootstrap/test/test_utils.dart
```diff
@@ -49,3 +53,15 @@ extension WidgetTesterX on WidgetTester {
...
 class _Dummy {} // ignore: unused_element
+
+/// Registers a mock [PageConfigService].
+///
+/// The [pages] argument will override the pages that are returned
+/// (empty by default) if provided.
+void setupMockPageConfig({Map<String, PageConfigEntry> pages = const {}}) {
+  final pageConfigService = MockPageConfigService();
+  registerMockService<PageConfigService>(pageConfigService);
+  when(pageConfigService.pages).thenReturn(pages);
```
d-loose, line 63: Misses a stub for `excludedPages` (error is thrown in `installer_test/initializes subiquity`, which somehow doesn't cause a test failure :confused:)

packages/ubuntu_provision/lib/src/services/page_config_service.freezed.dart
d-loose: Hm, I was certain I ran `pub upgrade`.. I'll check again :)

d-loose:
Looks very nice so far, great job!
Did you have a look at the test failures? I can't tell if they're related to recent yaru changes, since I can't reproduce them locally again..

Oh, and you seem to be mixing up github accounts causing the CLA check to fail.

d-loose:
> Did you have a look at the test failures? I can't tell if they're related to recent yaru changes, since I can't reproduce them locally again..

Sorted out flutter versions and dependency issues - test fail locally for me now. But it seems like they even fail on the `main` branch, so they might be related to other dependency changes. I'll investigate after lunch.

d-loose:
I suspect [this](https://github.com/canonical/ubuntu-flutter-plugins/pull/370) is what broke the tests.

spydon:
> Sorted out flutter versions and dependency issues - tests fail locally for me now. But it seems like they even fail on the `main` branch, so they might be related to other dependency changes. I'll investigate after lunch.

Ah, I just rebased on main, that must have been why I didn't see it previously.

d-loose (approved):
LGTM :+1:

PR #286, opened 2023-12-21
feat: Use page config files

spydon:
Since we (probably) don't want to handle defaults in another way than with the configs in the filesystem this PR introduces a fallback that reads the config from `assets/ubuntu-provision.[yaml|yml]`.

This PR also starts accepting images from assets to be used, since that most likely is what is going to be used by the fallback config.

The fallback config could look like this:
```yaml
pages:
  network:
    image: assets/images/network.svg
    title: My Network title
    visible: true
```

Note: this PR points to #285, so that PR should be merged first.


packages/ubuntu_provision/test/providers/page_images_test.dart
```diff
@@ -0,0 +1,70 @@
+// Write tests for page_images.dart
```
d-loose, line 1:
```suggestion
```

d-loose:
Seems like the integration tests are failing / not finishing due to the missing asset:
```
ERROR ubuntu_bootstrap: Unhandled exception
	FlutterError: Unable to load asset: "assets/ubuntu-provision.yaml".
The asset does not exist or has empty data.
```

spydon:
> Seems like the integration tests are failing / not finishing due to the missing asset:
> 
> ```
> ERROR ubuntu_bootstrap: Unhandled exception
> 	FlutterError: Unable to load asset: "assets/ubuntu-provision.yaml".
> The asset does not exist or has empty data.
> ```

Oops, seems like that file ended up in the next PR, I'll copy it over.

packages/ubuntu_bootstrap/assets/ubuntu-provision.yml
```diff
@@ -0,0 +1,33 @@
...
+    title: "Update available"
+    visible: true
+  source:
+    title: "Type of installation"
+    visible: true
+  secureBoot:
+    title: "Configure Secure Boot"
+    visible: true
+  storage:
+    title: "Ready to install"
```
d-loose, line 29: Even though the titles aren't used yet, the storage page should probably have a different one

spydon: Fixed! Probably a copy paste mistake

packages/ubuntu_bootstrap/lib/pages/source/source_wizard.dart
d-loose, line 1: Just out of curiosity: is there a particular reason to break the source wizard down into its pages?

d-loose (approved):
Great work, LGTM now! :+1:

packages/ubuntu_bootstrap/lib/pages/source/source_wizard.dart
spydon: It didn't really give anything for those two pages to be in a subwizard (if I didn't miss anything), it mostly just got confusing since it is using the index and length of its parent wizard.

PR #287, opened 2023-12-22
feat: Use page configs to determine which pages to show

spydon:
This reduces code duplication and makes it possible to control which pages that should be included or not from config.

### DIFF
--- a/packages/ubuntu_bootstrap/lib/installer.dart
+++ b/packages/ubuntu_bootstrap/lib/installer.dart
@@ -137,7 +137,7 @@ Future<void> runInstallerApp(
   tryRegisterService(UdevService.new);
   tryRegisterService(UrlLauncher.new);
 
-  final initialized = getService<SubiquityServer>().start(args: [
+  final initializeSubiquity = getService<SubiquityServer>().start(args: [
     if (options['dry-run-config'] != null)
       '--dry-run-config=${options['dry-run-config']}',
     if (options['machine-config'] != null)
@@ -147,7 +147,7 @@ Future<void> runInstallerApp(
     '--storage-version=2',
     '--dry-run-config=examples/dry-run-configs/tpm.yaml',
     ...options.rest,
-  ]).then(_initInstallerApp);
+  ]);
 
   await runZonedGuarded(() async {
     FlutterError.onError = (error) {
@@ -160,7 +160,9 @@ Future<void> runInstallerApp(
     final themeVariantService = getService<ThemeVariantService>();
     await themeVariantService.load();
     final themeVariant = themeVariantService.themeVariant;
-    await initialized;
+
+    final endpoint = await initializeSubiquity;
+    await _initInstallerApp(endpoint);
 
     runApp(ProviderScope(
       child: SlidesContext(
--- a/packages/ubuntu_bootstrap/lib/installer.dart
+++ b/packages/ubuntu_bootstrap/lib/installer.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 import 'dart:io';
 
-import 'package:args/args.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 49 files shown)


packages/ubuntu_bootstrap/lib/installer/installation_step.dart
```diff
@@ -0,0 +1,40 @@
...
+  source(SourcePage.new),
+  secureBoot(SecureBootPage.new),
+  storage(StorageWizard.new),
+  confirm(ConfirmPage.new);
+
+  const InstallationStep(this.pageFactory);
+
+  final ProvisioningPage Function() pageFactory;
+
+  WizardRoute? toRoute(BuildContext context, WidgetRef ref) {
```
d-loose, line 24: Does this have to be nullable?

packages/ubuntu_bootstrap/lib/pages/confirm/confirm_page.dart
```diff
@@ -4,13 +4,19 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
...
+class ConfirmPage extends ConsumerWidget with ProvisioningPage {
   const ConfirmPage({super.key});
 
-  static Future<bool> load(WidgetRef ref) {
+  @override
+  Future<bool> load(BuildContext context, WidgetRef ref) {
+    return ref.read(confirmModelProvider).init().then((_) => true);
+  }
+
+  static Future<bool> loadStatic(WidgetRef ref) {
```
d-loose, line 19: Is there a use case for this, or is it just a leftover from debugging?

spydon: Removed!

packages/ubuntu_bootstrap/lib/installer.dart
```diff
@@ -72,45 +83,57 @@ Future<void> runInstallerApp(
...
   tryRegisterService(JournalService.new);
-  tryRegisterService<KeyboardService>(
-      () => SubiquityKeyboardService(getService<SubiquityClient>()));
-  tryRegisterService<LocaleService>(
-      () => SubiquityLocaleService(getService<SubiquityClient>()));
-  tryRegisterService<NetworkService>(
-      () => SubiquityNetworkService(getService<SubiquityClient>()));
   tryRegisterService(
-      () => PageConfigService(config: tryGetService<ConfigService>()));
+    () => SubiquityKeyboardService(getService<SubiquityClient>()),
```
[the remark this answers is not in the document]
spydon: Ah, right, on all that specify a super type, or should we have the type parameters on all?

packages/ubuntu_bootstrap/lib/installer/installation_step.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:dartx/dartx.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:ubuntu_bootstrap/pages.dart';
+import 'package:ubuntu_provision/ubuntu_provision.dart';
+import 'package:ubuntu_wizard/ubuntu_wizard.dart';
+
+enum InstallationStep {
```
d-loose, line 8: This enum was previously only used for the progress indicator - this generalization nicely simplifies the route definitions :+1:.
However, there are some small issues with the indicator now - the second step is skipped and the number of steps changes:

[Screencast from 2024-01-02 14-36-32.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/113362648/b2e9e06d-a81d-4879-8271-995c015a1e85)

packages/ubuntu_bootstrap/lib/installer/installer_wizard.dart
```diff
@@ -68,76 +59,37 @@ class _InstallWizard extends ConsumerWidget {
...
-        Routes.confirm: WizardRoute(
-          builder: (_) => const ConfirmPage(),
-          userData: WizardRouteData(step: InstallationStep.storage.index),
-          onLoad: (_) => ConfirmPage.load(ref),
-        ),
+        ...preInstallRoutes,
         Routes.install: WizardRoute(
           builder: (_) => const InstallPage(),
-          onLoad: (_) => InstallPage.load(context, ref),
+          onLoad: (_) => const InstallPage().load(context, ref),
```
d-loose, line 84: Now that I'm seeing those two lines together - shouldn't we keep the `load` method static?

packages/ubuntu_bootstrap/lib/installer/installation_step.dart
spydon: How was that handled before when the second step was skipped? Where did the length come from? :thinking:

packages/ubuntu_bootstrap/lib/installer/installer_wizard.dart
spydon: Then we can't have the pages conform to an interface since static methods can't be overridden and then all pages need to be defined one by one, like it was before. Since they are const classes it is not any performance degradation here either, even though that wouldn't really have mattered.

packages/ubuntu_bootstrap/lib/installer/installation_step.dart
spydon: Aah right, because that step wasn't defined as a discrete step before.

PR #292, opened 2024-01-03
feat: add provd identity service

d-loose:
Replaces the DBus-based XDG identity service with one based on provd.

Since the backend doesn't maintain any state, the configured identity is saved in the service so that it can be retrieved later to launch the user's desktop session (via #298).

~~Some things that need to be figured out:~~

- ~~We probably don't want to create a user account immediately after the user presses 'next' on the identity page. Either we have to maintain a state in provd like subiquity does (in that case we should use similar methods, e.g. `setIdentity` instead of `createUser` and add an `apply` method that carries out the user creation) , or we need to handle this in the UI and keep a state in the client.~~
- ~~Depending on how we want to handle the previous point, we might not need a `getUser` method in provd. It's supposed to return the ephemeral user account information configured in the UI before the account is created.~~


packages/provd_client/lib/src/provd_user_client.dart
```diff
@@ -38,8 +39,12 @@ class ProvdUserClient {
...
   /// Validates the [username] and returns true if it is valid.
   Future<bool> validateUsername(String username) async {
-    final request = ValidateUsernameRequest(username: username);
-    final response = await _userClient.validateUsername(request);
-    return response.valid;
+    try {
+      final request = ValidateUsernameRequest(username: username);
+      final response = await _userClient.validateUsername(request);
+      return response.valid;
+    } on GrpcError catch (_) {
```
d-loose, line 46: So far provd only returns a boolean here and propagates additional information (i.e. why the username is invalid) as grpc errors. We could either generalize the grpc response (similar to how subiquity does this) or handle the different grpc error cases (once defined) here in the client and then wrap the result in an enum again.

packages/ubuntu_init/lib/src/services/provd_identity_service.dart
```diff
@@ -0,0 +1,41 @@
...
+
+  @override
+  Future<Identity> getIdentity() async => _identity;
+
+  @override
+  Future<void> setIdentity(Identity identity) async {
+    if (_identity == identity) return;
+    _identity = identity;
+    return apply();
+  }
```
d-loose, line 27: See comment at the beginning. If we want to maintain a state here, we should call `apply` separately at a later point after which we can't return to the identity page in the UI.

d-loose:
> We probably don't want to create a user account immediately after the user presses 'next' on the identity page. ...

Update: After some discussions about the first boot initialization step we could consider making the user creation irreversible, i.e. actually create the account when the user presses 'next' and remove the 'back' button on the following page.

Advanced users (especially those setting up a managed/enterprise machine) might want to open a terminal and perform actions using the new local user right away (while ubuntu-init is still running).

packages/provd_client/lib/src/provd_user_client.dart
d-loose: We decided to return an enum from the backend instead

packages/provd_client/lib/src/generated/accessibility.pb.dart
d-loose, line 1: Oh, didn't mean to delete all those ... Will restore them

matthew-hagemann (changes requested):
Hi Dennis,

I've run into a few issues with username validation in the new setup:

I am able to create a user from the reserved list (ie: root)
It looks like `provd` does fail, because if I click next, `ubuntu_init` gets stuck

Screenshot from 2024-01-18 12-27-47

Same issue for other reserved usernames. I also tried `gdm`

I also ran into an issue where valid usernames were rejected as being taken:

Screenshot from 2024-01-18 13-31-59

If it would be helpful, I can add some debug-logs to `provd` and push a new version up to help inspect.

d-loose:
Do you have the logs from ubuntu_init? I can't reproduce it so far.

matthew-hagemann (approved):
LGTM!

d-loose:
I assume all the issues you encountered were due to bugs in provd and/or your testing setup then?
In any case I'll only merge this once we're happy with the backend side :)

matthew-hagemann:
@d-loose Yea, I can't reproduce any of my issues after redoing my test infrastructure :slightly_smiling_face:
Cool beans! I'll send a PR out in a bit, just making some last changes for Didier :+1:

PR #293, opened 2024-01-04
feat: add postinst script to install provd from ppa

### LINKED ISSUE #252: Setup temporary pipeline for testing

d-loose:
~~@matthew-hagemann for this to work, you'll need to copy the package in the ppa to the 'noble' series (this can wait until you're back from the holidays, of course!).~~
Edit: done, thanks!

Should be good to go now

Close #252

### DIFF
--- a/snap/local/postinst.d/99_install_provd
+++ b/snap/local/postinst.d/99_install_provd
@@ -1,0 +1,13 @@
+#!/bin/bash
+TARGET=/target
+
+(
+cat << 'EOF'
+apt:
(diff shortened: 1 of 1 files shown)


matthew-hagemann (approved):
Looks good :slightly_smiling_face:

PR #298, opened 2024-01-11
feat(init): launch desktop session

### LINKED ISSUE #297: Launch user session after first boot initialization
After completing the first boot initialization, in which the first user account is created, we need to log in the new user and take them to the desktop.
Core desktop is currently doing this using [this](https://github.com/canonical/ubuntu-desktop-provision/pull/144/files#diff-29ff8861b552ad440711b934de53ffcf5c5668baa231cbd5d3aa0d1ca2bccb4a) bit of unmerged code. However, I'd propose to move this logic into `provd`, so that the UI can simply call something like a `LaunchUserSession()` method on the backend.

A big question that arises here is how this can be kept general enough so that it also works for non-GNOME based flavors. Would `provd` need to detect the desktop environment and then act accordingly?

d-loose:
After unsuccessfully trying to connect to GDM's DBus API from provd, we've decided to include this in the frontend instead.

This PR re-uses the code for the GDM service from #144 and adds tests. I've added a `launchDesktopSession()` method to the `InitModel` that can be called in the `onNext` callback of the last page in the init wizard.

Close #297 
UDENG-2026


spydon (approved):
Lgtm!

spydon:
Ah, didn't see that it was a draft. What's left?

d-loose:
Nothing left, just wanted to make sure the tests pass before marking it as ready for review.

sergio-costas (approved):
The code looks good. It calls OpenSession in the system bus, and then creates a local DBus connection specifying "messageBus: false", as required. Then, in the launchSession method connects the required signals and begin the user verification.

PR #325, opened 2024-01-17
feat(testing): install latest init snap

d-loose:
Simplify installation of init-snap


d-loose:
We'll get rid of the testing pipeline soon

PR #326, opened 2024-01-17
feat: rename ubuntu-desktop-installer snap to ubuntu-desktop-bootstrap

d-loose:
This will require some changes to the iso at some point.

For testing, you can still inject the ubuntu-desktop-bootstrap snap into an iso, boot it, quit the installer and remove it, then launch ubuntu-desktop-bootstrap instead. A bit cumbersome, but we should be able to get those things onto the iso soon..


matthew-hagemann (approved):
LGTM!

PR #242, opened 2023-12-01
feat: Create new `HorizontalPage` layout and use it for ubuntu_bootstrap

spydon:
[Screencast from 2024-01-12 14:00:45.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/744771/98c17436-2799-49f5-9248-c3bf6dcd94ae)

PR #327, opened 2024-01-17
fix: move page images to ubuntu_provision

d-loose:
Moves the page images to the shared `ubuntu_provision` package.
Also fixes #303 again - got lost in #242.

Edit: codecov seems to be having some issues again..


spydon (approved):
Lgtm

PR #328, opened 2024-01-17
feat: remove theme page

d-loose:
Since the theme page got removed from the new provision flows, it's not needed any longer.
For the moment the telemetry page becomes the last page, but we should consider adding a dedicated 'done' page for a nicer UX.


spydon (commented):
That's sad that this page was removed, was that really intentional from the design, or was it just overlooked?

d-loose:
Confirmed by @anasereijo

PR #331, opened 2024-01-18
feat: merge the latest changes from `main` into `bootstrap-snap`

d-loose:
Since we bumped the subiquity submodule on `main`, the subiquity integration test should fail on this PR. We'll have to update the subiquity dependencies in the snapcraft.yaml separately.


matthew-hagemann (approved):

d-loose:
> @d-loose is the failing subiquity integration test expected?

Yes, see original comment ;)

PR #333, opened 2024-01-18
chore(provd): remove hello service

matthew-hagemann:
Removing the hello world gRPC service now that we have an implemented service to test against.


d-loose (approved):
Thanks, LGTM :+1:

PR #334, opened 2024-01-18
User dbus path fix

matthew-hagemann:


d-loose (approved):
Thanks! :+1:
It'd be awesome if copy/paste errors like that would cause a test failure :)

PR #335, opened 2024-01-19
fix(provd): multiple code review fixes

matthew-hagemann:
- Updated tests to rely on mocks for testing error cases, not additional dbus objects
- Password hashing via cgo to align with how GIS hashes passwords
- Added checks to see if dbus objects are available when create a new user client (did require tests from daemon / manager to be updated so that ping mocks were available on them aswell when the new user client is initialized)
- CreateUser attempts to rollback setting the `statichostname` and delete the created user
- Reserved usernames read from `/usr/share/base-passwd/*`
- Test cases reformatting to make tests more readable
- Added testutils methods for reusable testing utils
- Increased test coverage


d-loose:
Looks great so far!
However, libaccountsservice uses `crypt_gensalt` (which supports yescrypt) from [libxcrypt](https://github.com/besser82/libxcrypt) when available and only uses the custom sha512crypt implementation as a fallback. See https://gitlab.freedesktop.org/accountsservice/accountsservice/-/blob/main/src/libaccountsservice/act-user.c?ref_type=heads#L1753-1789
Is that something we could do in CGo as well?

matthew-hagemann:
> Hey! matthew-hagemann has not signed the Canonical CLA which is required to get this contribution merged on this project.
> 
> Please head over to https://ubuntu.com/legal/contributors to read more about it.

:raised_eyebrow:

matthew-hagemann:
Sorry, I've just realized I have once again missed the bullet point on my list that says "READ RESERVED NAMES FROM SYSTEM!!!!"

I will add that now :sweat_smile:

d-loose:
Not sure what the underlying issue is, but when I try to create a user that already exists, I still get
```
ERROR:
  Code: Internal
  Message: grpc: error while marshaling: proto: Marshal called with nil
```
as a response and
```
2024/01/22 17:59:02 ERROR: [core] [Server #1] grpc: server failed to encode response: rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil
```
in provd's logs.

provd/internal/services/manager_test.go
```diff
@@ -64,9 +69,84 @@ func requireEqualServices(t *testing.T, want, got map[string]grpc.ServiceInfo) {
...
+
 func TestMain(m *testing.M) {
 	testutils.InstallUpdateFlag()
 	flag.Parse()
 
+	defer testutils.StartLocalSystemBus()()
+
+	conn, err := dbus.SystemBusPrivate()
+	if err != nil {
+		slog.Error("Setup: can't get a private system bus: %v", err)
```
d-loose, line 92: Noticed while debugging this that `slog.Error()` doesn't support format strings

provd/internal/services/user/user_test.go
```diff
@@ -52,81 +152,55 @@ func TestCreateUser(t *testing.T) {
...
-
-		accountsEndpoint string
-		hostnameEndpoint string
 	}{
 		"Successfully creates a user": {
-			realName:  "Ubuntu",
-			username:  "ubuntu",
-			password:  "password",
-			hostname:  "ubuntu",
+			realName:  "ok",
```
didrocks, line 158: So, as you can see, most of the tests have:

`
realName:  "ok",
username:  "ok",
password:  "ok",
hostname:  "ok",
```

which is tedious :)

I suggest that we use default values, don’t set them, then in the subtests:

```
if tc.username == "" {
   tc.username = "ok` // (I think a better name can be used too)
}
```

Then, you are going to tell me "yes, but what if I want the default value for that element (an empty string in that case", and this is where I answer "no problem, just have a convention". What we do in enterprise desktop/WSL is that we picked "-". That means "empty" (which is generally just one test case).

You then end up with:
```
if tc.username == "" {
   tc.username = "ok` // (I think a better name can be used too)
} else if tc.username == "-" {
   tc.username = ""
}
```

And if you want to be fancy and add a third one, you can even use a switch!

That way, you are going to remove many lines in the test case definition, and you can then ordering them to be more compact (as you have fewer fields, only you can have everything in one line for most of there), and as you are going to deviate from the default like "when hostname is empty", you set this hostname to empty as the **first** field to assign in your named struct for that test. That way, the test case become:

`"Error when hostname is empty": { hostname:  "ubuntu", wantErr:   true}`

as everything else is not relevant and then covered by our default set in the subtests.

Then, you can organize your tests cases with:

```
// Positive tests, the easiest first
< line break>
// Positive tests, but special cases or special behaviours, you can even have many sections delimited by lines break with a comment to illustrate groups of related special cases
< line break>
// Error cases
```

Some example of this:

You can see the struct definition is separated (as you do!): https://github.com/ubuntu/authd/blob/main/internal/services/pam/pam_test.go#L248

The tests cases are then grouped and most of time, one liner, very easy to read and see only the changed parameters impacted this use case: https://github.com/ubuntu/authd/blob/main/internal/services/pam/pam_test.go#L260

We do have some defaults for some of those values to avoid repetition: https://github.com/ubuntu/authd/blob/main/internal/services/pam/pam_test.go#L285

Oh, and while I am at it, nitpick, but `username, hostname, but realName` -> I think we can drop the capital N then :)

File: provd/internal/services/user/user_test.go [hunk 3]
```diff
@@ -34,13 +38,109 @@ func TestEmptyCreateUserRequest(t *testing.T) {
...
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(testutils.StartLocalSystemBus())
+			bus := testutils.NewDbusConn(t)
+
+			var opts []user.Option
+			v := reflect.ValueOf(tc)
+			for i := 0; i < v.NumField(); i++ {
+				if v.Field(i).String() != "" {
```
didrocks, line 124: This is smart, but not clever :p in the sense that it will be quite hard to maintain. I’m sure you can achieve something very similar, in approximately the same amount of line of codes with what I wrote in the next comment (if none are added meanwhile :p just check for the big comments on test cases)

provd/internal/services/user/user_test.go
```diff
@@ -34,13 +38,109 @@ func TestEmptyCreateUserRequest(t *testing.T) {
...
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Cleanup(testutils.StartLocalSystemBus())
+
+			var opts []user.Option
+			if tc.invalidAccounts {
+				opts = append(opts, user.WithAccountsPrefix("invalid"))
```
didrocks, line 77: isn’t this test the same than one below?

provd/internal/services/user/user.go
```diff
@@ -186,51 +201,83 @@ func (s *Service) ValidateUsername(ctx context.Context, req *pb.ValidateUsername
...
+// Function to scan a file for the username.
+func scanFile(file *os.File, username string) (bool, error) {
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(line, "#") || line == "" {
+			continue
+		}
+		// Extract the username from the line
+		fields := strings.SplitN(line, ":", 2)
```
didrocks, line 277: I would error out if the file does not have the expected format, So failing early, continuing if not found, and the "happy path" is one tab more left, with the return.

provd/internal/services/user/user.go
```diff
@@ -28,78 +26,85 @@ const (
...
+		}
+
+		// Rollback hostname
+		err = s.hostname.Call(consts.DbusHostnamePrefix+".SetStaticHostname", 0, currentHostname, false).Err
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to rollback hostname: %v", err))
+		}
+		// Delete user
+		if userID == 0 {
+			return
```
didrocks, line 100: when you run coverage profile, is this line covered? (I didn’t see any CreateUser with 0 as a parameter which should trigger the rollback here and this stop call)

provd/internal/services/user/user.go
```diff
@@ -28,78 +26,85 @@ const (
...
+
+	defer func() {
+		if err == nil {
+			return
+		}
+
+		// Rollback hostname
+		err = s.hostname.Call(consts.DbusHostnamePrefix+".SetStaticHostname", 0, currentHostname, false).Err
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to rollback hostname: %v", err))
```
didrocks, line 96: I tried to apply this when running a service:
- everything that stops the service is then an error and should be logged as an error
- everything that don’t directly stops the services, but is important to the user is then just a warning (otherwise, we would have stopped).

Here, in the recovery, not resetting the hostname is a best effort as a rollback item, but failing to do so is not exactly what makes us stop (it’s the previous error that did), so I would go for just a warning in that case.

Also, you can avoid an undeed warning confusing the error by only trying to rollback if the currentHostame != from the one you request to change it too? (we will still have cases then when we reset even if we didn’t change it, but it’s a little bit more constraints still)

provd/internal/services/user/user.go
```diff
@@ -186,51 +201,83 @@ func (s *Service) ValidateUsername(ctx context.Context, req *pb.ValidateUsername
 		return &pb.ValidateUsernameResponse{UsernameValidation: pb.UsernameValidation_TOO_LONG}, nil
 	}
 
-	// Check if username is in reserved list
-	// Read line by line to avoid loading the whole file into memory
-	file, err := reservedUsernamesFS.Open("reserved-usernames")
+	// Open the passwd.master file
+	passwdFile, err := os.Open("/usr/share/base-passwd/passwd.master")
```
didrocks, line 205: How do you mock that in tests?

Trigger cases when there is no file, files with incorrect parsing and so on?

didrocks (dismissed):
Ok, This is great progress! Congrats on the dbus part, this is heading on the correct direction and so alien to the rest of the code :)

I did write some comments on what I immediately spot, feel free to read and tell me if there is anything that isn’t clear to you.

Do not forget to use coverage to guide some of the decisions and testing you do! I find it personnally really helping myself.

provd/internal/services/user/user_test.go
```diff
@@ -298,56 +342,71 @@ func newUserClient(t *testing.T, accountsEndpoint string, hostnameEndpoint strin
...
-		return "", dbus.NewError("org.freedesktop.Accounts.Error.Failed", []interface{}{"error"})
+	if name == "error" {
+		return "", dbus.NewError("org.strange.error", []interface{}{"error"})
 	}
-	return "/org/freedesktop/Accounts/Usernoerror", nil
+	// This is used to also include the username in the error log, used to determine its an error because a user wasn't found.
+	if name != "ok" {
+		return "", dbus.NewError("org.freedesktop.Accounts.Error.Failed", []interface{}{"Your name was: " + name})
+	}
+	return "/org/freedesktop/Accounts/Userok", nil
```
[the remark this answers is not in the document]
matthew-hagemann: I hardcoded the returned user because that is the path of the user I create when I export the dbus mocks at the start of TestMain(). So in all the positive cases, they are interacting with the same pre-exported user object mock?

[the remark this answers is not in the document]
matthew-hagemann: Path is actually dependent on the UserID I see. So Creating a second user on my PC with username matt2 will return `/org/freedesktop/Accounts/User1001`

provd/internal/services/user/user.go
matthew-hagemann: We could have the filepaths configurable as options on the User service struct?

didrocks: exactly!

provd/internal/services/user/user.go
matthew-hagemann: It is caught in the testcases when we fail before a user is created, so it gets covered in the tests that cover the input validation

matthew-hagemann:
Hi @didrocks :smile: I have made the majority of the changes you requested. The only thing I am unsure of is how to approach testing different cases on `Accounts.DeleteUser`, `hostname1.Get` and `User.Get`. While these functions take in parameters we could switch on to simulate failure cases, they are not easy for us to influence. 

DeleteUser takes in a Uid, but in the code impl, this is gained through calling User.Get("Uid") on the mocked User dbus object, which itself is difficult to pass conditions to in order to influence what the mock returns to you other than a static pre-set value:

```
func (u userdbus) Get(interfaceName string, propertyName string) (interface{}, *dbus.Error) {
        ...
	return uint64(1), nil
}

func (a accountsdbus) DeleteUser(userID uint32) *dbus.Error {
        ...
	return nil
}
```

The only option I can really think of is creating multiple mocked dbus user objects, that will return different user id's based on what we set on creating of the mock, and then influence how delete works depending on what ID is called? Similar story for getting hostnames. WDYT?

didrocks:
> The only option I can really think of is creating multiple mocked dbus user objects, that will return different user id's based on what we set on creating of the mock, and then influence how delete works depending on what ID is called? Similar story for getting hostnames. WDYT?

That would have been my suggestion too. You can imagine creating an user "fail-deletion", which returns a very recognizable uid (like 9999 for instance), which will make the deleteuser failing. Similar to the hostname.

As the mock is self-constaints and the methods are one next to each other in the same file, I think the relationship will be really clear (you can even create a `const` in that file for the uid and have a recognizable name like `uidUserDeletionFails`)

matthew-hagemann:
Perfect! Thanks :smile:

matthew-hagemann:
Hi @didrocks! I have made the remaining changes. I've double checked my handwritten list and your comments, so I hope I haven't missed anything! Let me know what you think :slightly_smiling_face:

d-loose (approved):
Tested everything again - error messages look good password hashes are valid and even yescrypt is supported :tada:

LGTM :+1:

PR #340, opened 2024-01-22
feat(init): disable back buttons on certain pages

### LINKED ISSUE #305: Disable 'previous' buttons on certain pages in the init wizard
Some steps during first boot initialization (like creating the first user account) will be carried out immediately and are irreversible / cannot be performed twice. Hence we'll need to turn certain parts of the init wizard into a one-way street by removing the 'previous' buttons.

d-loose:
As some actions in the init wizard are done immediately and aren't trivially reversible, some of the wizard steps shouldn't allow the user to navigate back. I think the simplest solution is to explicitly set this as an individual property for each page in the wizard.

Close #305
UDENG-2061


spydon (approved):
Lgtm!

PR #341, opened 2024-01-22
refactor: `NextWizardButton` and `PreviousWizardButton` widgets

spydon:
This refactor (which was just supposed to be a fix), does the following:
 - Breaks out `WizardButton.next` and `WizardButton.previous` to discrete widgets
 - Fixes so that you can't multi-tap buttons
 - Fixes race conditions that locked the wizard
 - Removed the use of `FutureBuilder` for the internal loading state
 - Sets the defaults in the constructor so that they are more easily read from the user of the widget
 - Renames `onBack` to `onExecute` in `NextWizardButton` (to clarify that it runs when `execute` runs)
 - Hopefully makes the code a bit easier to read by removing nested short-hand ifs etc


d-loose:
Thanks, looks like a big improvement! I think you might need to `pumpAndSettle` after tapping the 'next' button in the failing tests to resolve the timer issue.
I'll have a closer look at the changes after the meeting.

packages/ubuntu_wizard/lib/src/wizard_button.dart
```diff
@@ -16,131 +16,49 @@ final _noAnimation = Listenable.merge([]);
...
   /// Determines whether the button is highlighted (green);
   ///
   /// The default value is `false`.
-  final bool? highlighted;
+  final bool highlighted;
 
   /// Determines whether the button is enabled.
   ///
-  /// The default value is `false`. If not specified, the button is implicitly
+  /// The default value is `true`. If not specified, the button is implicitly
```
d-loose, line 55: Just double-checking that this is intended. Has the logic or the semantics changed here?
If the default value for `enabled` is `true` and `onActivated` is `null`, will the button be disabled or enabled?

d-loose:
For me the 'Next' button briefly changes to 'Done' and then back to 'Next':

[Screencast from 2024-01-23 15-02-32.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/113362648/fd4531f4-78ed-4f5d-8e9a-fc27f159fa88)

packages/ubuntu_wizard/lib/src/wizard_button.dart
spydon: It will be disabled, but the field `enabled` defaults to true. The semantics have changed.
You can see here how it defaulted to true before too:
https://github.com/canonical/ubuntu-desktop-provision/pull/341/files#diff-8ea315d2174db0c87c77ccc3b4f004d1b41357277a429bff6eba7a7b636028a8L95

spydon:
> For me the 'Next' button briefly changes to 'Done' and then back to 'Next':

Baah, I thought I fixed that issue. I'll look into it!

packages/ubuntu_wizard/lib/src/wizard_button.dart
```diff
@@ -152,52 +70,206 @@ class WizardButton extends StatefulWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final wizard = Wizard.maybeOf(context);
+    final rootWizard = Wizard.maybeOf(context, root: true);
+    final routeData =
+        (wizard?.routeData ?? rootWizard?.routeData) as WizardRouteData?;
+    final hasNext = [
+      routeData?.hasNext,
+      wizard?.hasNext,
+      rootWizard?.hasNext,
```
spydon, line 242: It seems like these are all eventually non-null, do you know if this is always the case @d-loose?

I don't understand why the previous button implementation didn't show the done label before, because if I copy in that logic here now it still shows the done label. Anyhow, this seems to work like it should.

spydon: Ah, `routeData` can of course be null sometimes, fixed that.

d-loose: Right, `routeData` is entirely optional.
`wizard` and `rootWizard` are only nullable because `Wizard.maybeOf` is used instead of `Wizard.of` here. Not sure if there's a sensible case where either of those can be null.

packages/ubuntu_wizard/lib/src/wizard_button.dart
d-loose: ok, thanks that makes sense. Just a bit slow on evaluating boolean expressions in my head today :)

spydon: Oh boy do I know that feeling, definitely been having that today too 😄

d-loose (approved):
Awesome, that works really well now - much better than it was before!

packages/ubuntu_wizard/lib/src/wizard_button.dart
d-loose: I'm just realizing that this broke #340. `routeData` should always take precedence if provided, otherwise it doesn't serve any purpose.

PR #345, opened 2024-01-24
fix: Support window sized down to 700x560

spydon:
With a minimum display size of 800x600 the window needs to be able to be 700x560, since the top bar and the dock has to be accounted for.

With this PR extra horizontal padding and scroll views are introduced to make it work on the minimum size.
The default size is still 960x680 though, and if the screen is smaller than that the window maximizes itself.

### DIFF
--- a/packages/ubuntu_bootstrap/lib/pages/source/source_page.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/source/source_page.dart
@@ -31,54 +31,58 @@ class SourcePage extends ConsumerWidget with ProvisioningPage {
       name: 'source',
       windowTitle: lang.updatesOtherSoftwarePageTitle,
       title: lang.updatesOtherSoftwarePageDescription,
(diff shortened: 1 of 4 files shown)


d-loose:
Thanks, works really well!

Two tiny nitpicks:

Text is still cut-off on the source page, even with the default window size:
Screenshot from 2024-01-24 17-18-44

Text can run into the scrollbar on the network page:
Screenshot from 2024-01-24 17-16-45

spydon:
> Text is still cut-off on the source page, even with the default window size:

Yeah, that is due to this in yaru_widgets:
https://github.com/ubuntu/yaru_widgets.dart/blob/main/lib/src/widgets/yaru_toggle_button.dart#L84

> Text can run into the scrollbar on the network page:

Strange that Flutter doesn't handle that automatically

d-loose (approved):
LGTM :+1:

Issue #204, opened 2023-11-17
Create a new 'welcome' UI


d-loose:
In the new provisioning flow, we're planning to have a separate 'welcome' stage that is entered after the first boot initialization (currently gnome-initial-setup, to be replaced by ubuntu-desktop-init) finishes. It's supposed to welcome the user, display release note, point them to documentation resources and lead them to other applications like the app center.

d-loose:
Since we've decided to simplify the new flow and not have a separate 'welcome' wizard after all, I'm closing this.

d-loose:
Re-opening this, since plans have changed again :)

PR #346, opened 2024-01-24
feat: simplify page configuration

d-loose:
While trying to implement a separate 'welcome' wizard (#204) I ran into various issues with the current page configuration setup and decided to simplify it a bit.

* Consistently use kebab case in config files
* Always include all pages defined in `InstallationStep` or `InitStep` in the wizard
* Hide individual pages using the wizard's route `predicate` again
* Use fixed length page indicator (assuming all pages are visible). If pages are hidden via a config file setting (either in ubuntu-provision.yaml, or through `interactive-sections` in the autoinstall.yaml passed to subiquity) the indicator will simply skip them. If necessary we can make the page indicator adapt to the set of visible pages later.
* Use a single default config file in `ubuntu_provision/assets` for all wizards.
* Remove `pageConfigProvider` and rely on the `InstallationModel` and `InitModel` that talk to the `PageConfigService` instead
* Remove `includedPages` from `PageConfigService` as it caused some double bookkeeping

UDENG-1760


packages/ubuntu_provision/test/test_utils.dart
```diff
@@ -84,3 +86,31 @@ class _Dummy {} // ignore: unused_element
   MockSpec<SvgPicture>(),
 ])
 class _Dummy2 {} // ignore: unused_element
+
+class FakeAssetBundle extends CachingAssetBundle {
```
d-loose, line 90: This is the same as in the xdg keyboard service test. I didn't move it into a shared package like ubuntu_utils and just copied it instead, since we'll hopefully get rid of the xdg keyboard service soon.

packages/ubuntu_provision/assets/ubuntu-provision.yml
```diff
@@ -1,6 +1,39 @@
-# This file is only to be used for tests
-scope:
-  test:
+pages:
+  locale:
     title: "Welcome"
-    image: "mascot.png"
-    visible: true
\ No newline at end of file
+    image: "packages/ubuntu_provision/assets/mascot.png"
```
d-loose, line 4: Maybe it'd also be worth making `packages/ubuntu_provision` the default for asset images, so that we can just specify `image: assets/mascot.png` here?

spydon (approved):
Lgtm, the page indicator will still show the excluded pages now right?
Should the `ModalRoute.of(context)` soludion going to replace `name` in this PR too, or will that be a follow up?

d-loose:
> Lgtm, the page indicator will still show the excluded pages now right?

Yes, now in both cases again if pages are excluded by either by subiquity or in the ubuntu-provision.yaml. Before the indicator would only adjust its length if pages were excluded by the config, but still show all pages when they were marked as non-interactive by subiquity. I think it'd be better to find a solution that works for both cases if necessary.

> Should the `ModalRoute.of(context)` soludion going to replace `name` in this PR too, or will that be a follow up?
Ah right, I can do that here as well!

packages/ubuntu_provision/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   args: ^2.4.2
   async: ^2.11.0
   collection: ^1.17.0
+  dartx: ^1.2.0
```
spydon, line 14: Is it really worth adding a dependency for only `removePrefix`? :sweat_smile: 
Something like `replaceFirst(r'^\/', '')` can be used instead of `removePrefix('/');`, or simply `replaceFirst('/', '');` since all will start with `/`.

d-loose: Hm, we got it in ubuntu_bootstrap already and it's listed in the melos.yaml as well, so I guess it doesn't hurt? But I'm happy to adopt your solution if that's cleaner

spydon: It doesn't seem to be used in ubuntu_bootstrap anymore, it's just in the pubspec, so probably better to just remove it everywhere?

d-loose: It's used in various places, but in all cases there are replacements from `dart:core` or `collection`

d-loose: Unexpected subtle differences between `elementAtOrNull` from `collection` vs `dartx` :/
It's probably a good idea to remove dartx :+1:

spydon (approved):
Nice!

PR #352, opened 2024-01-26
feat: cleanup desktop service

### LINKED ISSUE #338: `DesktopService` exceptions hinder subsequent service initializations
Discovered [here](https://bugs.launchpad.net/ubuntustudio-system-installer/+bug/2049675).
If any exceptions are thrown in the `DesktopService` initialization (ubuntu_bootstrap), the following service registrations and initializations are never executed, causing the wizard to crash.

We should:
* handle and log errors/exceptions from the `DesktopService`
* refactor the service initialization code
* implement the `DesktopService` for flavors that aren't based on GNOME

d-loose:
* Don't register `DesktopService` in debug mode to not constantly mess up the developer's power settings
* Refactor `DesktopService` and catch/log all potential errors from non-existing dconf paths or DBus services so that flavors don't run into problems. We'll probably need to add additional `DesktopService` implementations at some point.
 This should fix https://bugs.launchpad.net/ubuntustudio-system-installer/+bug/2049675.
 I'd cherry-pick this into `archive/main`.
* Make sure the application ID in the Gnome Session `Inhibit()` call matches the one defined in CMakeLists.txt

Fix #338

### DIFF
--- a/packages/ubuntu_bootstrap/lib/installer.dart
+++ b/packages/ubuntu_bootstrap/lib/installer.dart
@@ -89,7 +89,7 @@ Future<void> runInstallerApp(
   tryRegisterServiceInstance<ArgResults>(options);
   tryRegisterService<ConfigService>(
       () => ConfigService(path: options['config'] as String?));
-  tryRegisterService<DesktopService>(GnomeService.new);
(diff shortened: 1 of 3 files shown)


spydon (approved):
Lgtm

PR #366, opened 2024-01-31
feat: add a11y service interface and gnome implementation

### LINKED ISSUE #362: Implement accessibility model for ubuntu-bootstrap
In order for the new accessibility page to work in the live system (running ubuntu-bootstrap), we need to create a Dart service that changes the corresponding dconf settings, as subiquity doesn't operate at that.

d-loose:
The first part is a general a11y service interface that includes all of the settings visible on the accessibility page so far and can be easily extended to include more.
The second part is a dconf based implementation for gnome that can be used in stage 3 if we decide to do so, despite the fact that this is possible through gnome's own a11y menu already.

Close #362
UDENG-2203


spydon (approved):
Clean! Looks good to me :)

PR #370, opened 2024-01-31
feat: reintegrate identity, active directory and timezone pages into ubuntu bootstrap

### LINKED ISSUE #369: Bring back user creation and timezone to ubuntu bootstrap
We've decided to re-include the user creation page as an optional step into ubuntu bootstrap (stage 3). This will enable flavors and possibly the standard desktop installer to create a user during the installation (like in the 23.10 installer) and completely skip the first boot init wizard (stage 4), while still supporting the new flow (no user creation in stage 3, first boot init with user creation in stage 4) for the OEM use case.
This also means that the timezone page needs to be re-included in ubuntu bootstrap, as that step would otherwise be missing if stage 4 is skipped.

Essentially all this amounts to reverting #212.
See also #199.

d-loose:
Adds the identity, active directory and timezone pages back into ubuntu bootstrap (they're now shown immediately before the confirmation page). I've removed the `FakeSystemService` we used in ubuntu init to test user creation and restored the original behavior of `expectIdentity`.

We'll need to remove the [postinst script](https://github.com/canonical/ubuntu-desktop-provision/blob/44dff1bea18dcc384d7f7d268946773ccd480e19/snap/local/postinst.d/99_disable_user_creation) that disables user creation from the snap and then find a solution for the OEM case where the identity page will be hidden and no user account should be created. I'll open separate issues for those things after merging this.

Looks like this now:

[Screencast from 2024-01-31 18-39-30.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/113362648/4d2f6f99-4c46-415f-a568-7167b11bbd61)

@anasereijo I suspect we might want to move the timezone page further to the beginning here, next to the locale/keyboard, right?

I noticed we didn't have a privacy page before the timezone page in the 23.10 installer, so that means location services are enabled in the live system by default?

P.S. @spydon the numbering in the screenshot test is all over the place now :D let's sort this out once we revive the screenshot repo.

Fix #369
UDENG-2218

### DIFF
--- a/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
+++ b/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
@@ -325,7 +325,57 @@ Future<void> main() async {
     await tester.pumpAndSettle();
 
     await tester.testConfirmPage(
-      screenshot: '$currentThemeName/10.confirm',
+      screenshot: '$currentThemeName/09.confirm',
+    );
+  }, variant: themeVariant);
+
+  testWidgets('10.timezone', (tester) async {
+    await tester.runApp(() => runInstallerApp([], theme: currentTheme));
+    await tester.pumpAndSettle();
+
+    await tester.jumpToPage(Routes.timezone);
+    await tester.pumpAndSettle();
+
+    await tester.testTimezonePage(
+      screenshot: '$currentThemeName/10.timezone',
+    );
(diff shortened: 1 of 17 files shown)


spydon (approved):
Lgtm :)

d-loose:
Successfully tested this on an iso - I'll merge this and open the follow-up issues I mentioned.

PR #371, opened 2024-02-01
Locale impl

### LINKED ISSUE #312: locale service impl
Implement the locale grpc service

matthew-hagemann:
Closes  #312


provd/example.config.yaml
```diff
@@ -1,2 +1,2 @@
 Paths:
-  Socket: "/run/test.sock"
+  Socket: "/home/matt/test.sock"
```
d-loose, line 2: Was that intentional? I was wondering why my local test setup broke and found this :D

d-loose (approved):
Tested it using `grpcurl` and using the dart client I'm generating/writing now - works without any issues! :+1:
Possible future improvement, as discussed: return a status from `SetLocale` if the user needs to wait for locale generation to finish. For now we can display a generic loading spinner in the UI.

matthew-hagemann:
Perfect! Thanks Dennis. I'll wait for your impl and we can make sure we are happy end to end before we merge :slightly_smiling_face:

provd/example.config.yaml
matthew-hagemann: Haha, no, that was not intentional :sweat_smile: I'll revert that back!

d-loose: I did the [same](https://github.com/canonical/ubuntu-desktop-provision/compare/a28f5bbc95b932698c3ebee26f94bbc457ec48f4..0bbaa6168c5b3ddedcd039e1091755f890cc9dd6) now.. :grimacing:

Issue #363, opened 2024-01-29
Remove remaining services from the init wizard


d-loose:
The init wizard should completely rely on the `provd` backend to carry out all of the necessary actions (probably with the exception of launching the desktop session). The existing Dart XDG services need to be removed and replaced with `provd`-based services (or temporary fake services).

PR #373, opened 2024-02-01
feat(init): migrate to provd locale service

d-loose:
Replaces the xdg locale service with a provd-based service, ready for use with #371.

Ref #363 
UDENG-2204


spydon (approved):
Lgtm

matthew-hagemann (approved):
Gave it a run, all seem good! I see what you mean regarding the wait for the locale to generate, but as you suggested, we can look not blocking on that in the future :+1:

PR #374, opened 2024-02-01
fix: wifi_views to not be flexible in a non-constrained context

spydon:
Before the `WifiView` crashed when another radio button was chosen due to how flex was set up on it.


d-loose (approved):
LGTM :+1:

Issue #315, opened 2024-01-15
keyboard service impl


matthew-hagemann:
Implement the keyboard grpc service

PR #381, opened 2024-02-02
feat(init): migrate to provd keyboard service

d-loose:
Replaces the xdg keyboard service with a provd-base one, ready for use with #315.
I've cherry-picked the proto changes and fixed the typos here - will rebase once #315 is updated.

Ref: #363 
UDENG-2204


packages/ubuntu_init/lib/src/services/provd_keyboard_service.dart
```diff
@@ -0,0 +1,60 @@
...
+extension on provd.KeyboardLayout {
+  KeyboardLayout fromProvd() => KeyboardLayout(
+        code: code,
+        name: name,
+        variants: variants.map((e) => e.fromProvd()).toList(),
+      );
+}
+
+extension on provd.KeyboardVariant {
+  KeyboardVariant fromProvd() => KeyboardVariant(
```
spydon, line 56: Does this method do anything? Isn't it the same data as it was before?

spydon (approved):
Lgtm

It's a bit sad that we get all the duplicated class names, but probably nothing we can do about that

packages/ubuntu_init/lib/src/services/provd_keyboard_service.dart
d-loose: It converts a `provd.KeyboardVariant` to a `KeyboardVariant` from subiquity's types

PR #388, opened 2024-02-05
feat(init): migrate to provd based timezone service

d-loose:
Replaces the xdg timezone service with a provd-based one.

@matthew-hagemann let me know if the protos are sufficient, or if we want to treat a special case explicitly: subiquity automatically determines the timezone if you call `setTimezone('geoip')` - we should probably support that too.

Ref: #363 
UDENG-2204


matthew-hagemann:
@d-loose Timezone protos look good to me!

spydon (approved):
Lgtm!

PR #391, opened 2024-02-06
feat: Merge custom whitelabel file with the default

### LINKED ISSUE #377: whitelabel.yaml: merge externally provided config with defaults from assets
Currently we're taking default configuration settings from  the whitelabel.yaml in `packages/ubuntu_provision/assets`. However, if an external whitelabel.yaml is found, we use this instead and ignore the one from the assets.
It'd be a lot cleaner if we always read the default values from the asset file and then use the external config file to override them.

spydon:
Closes #377

### DIFF
--- a/packages/ubuntu_provision/lib/src/services/config_service.dart
+++ b/packages/ubuntu_provision/lib/src/services/config_service.dart
@@ -6,6 +6,7 @@ import 'package:flutter/services.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:ubuntu_utils/ubuntu_utils.dart';
 import 'package:xdg_directories/xdg_directories.dart' as xdg;
 import 'package:yaml/yaml.dart';
 
@@ -125,8 +126,8 @@ class ConfigService {
     Map<String, dynamic> defaultConfig,
     Map<String, dynamic>? customConfig,
   ) {
-    if (customConfig == null) return defaultConfig;
-    final result = Map<String, dynamic>.from(defaultConfig);
+    final result = defaultConfig.toMap();
+    if (customConfig == null) return result;
     customConfig.forEach((key, value) {
       if (value is Map && result[key] is Map) {
         result[key] = mergeConfig(
@@ -137,7 +138,7 @@ class ConfigService {
         result[key] = value;
       }
     });
-    return result;
+    return result.toMap();
   }
 }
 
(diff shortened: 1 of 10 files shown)


d-loose (commented):
Thanks, looks good so far! `mergeConfig()` has issues with `CastMap` and `YamlMap` though, so it doesn't (yet) work in practice. Maybe it could be helpful to add a test case with a custom file and a fake asset (like in the 'yaml' test).

spydon:
> Thanks, looks good so far! `mergeConfig()` has issues with `CastMap` and `YamlMap` though, so it doesn't (yet) work in practice. Maybe it could be helpful to add a test case with a custom file and a fake asset (like in the 'yaml' test).

Not sure I understand, `mergeConfig` should only need to handle nested `Map<String, dynamic>` maps, since they are already casted once that method is called? I probably misunderstand something in your comment though.

d-loose:
When I test this with an external config containing
```yaml
theme:
  accent-color: "#00ff00"
pages:
  accessibility:
    visible: false
```
the default config gets completely overridden.

At a breakpoint in `load()` I see:
```
-> defaultConfig.runtimeType
> Type (CastMap<dynamic, dynamic, String, dynamic>)
-> defaultConfig['pages'].runtimeType
> Type (YamlMap)
```
So the entries are still `YamlMaps` in practice and merging the nested maps fails.

spydon:
@d-loose aah, you're correct! It should be fixed now, and the file path in the test goes away in #392

d-loose:
Thanks! I think there's one final issue: 
The `PageConfigService` expects a `YamlMap`
```dart
  Future<void> load() async {
    final pageConfig = PageConfig.fromJson({
      'pages': Map<String, dynamic>.from(
        (await _config!.get<YamlMap>('pages'))?.value.cast() ?? {},
...
```
but now only gets one if the default config remains unmodified. Otherwise it receives the merged map (`Map<String,dynamic>`) it fails with:
```
_TypeError: type '_Map<String, dynamic>' is not a subtype of type 'YamlMap?' in type cast
#0      ConfigService.get (package:ubuntu_provision/src/services/config_service.dart:37:28)
#1      PageConfigService.load (package:ubuntu_provision/src/services/page_config_service.dart:31:25)
#2      _initInstallerApp (package:ubuntu_bootstrap/installer.dart:238:37)
#3      runInstallerApp.<anonymous closure> (package:ubuntu_bootstrap/installer.dart:176:11)
<asynchronous suspension>
#4      runInstallerApp (package:ubuntu_bootstrap/installer.dart:164:3)
<asynchronous suspension>
```
And this is a case we don't have a test for.

d-loose (approved):
Awesome, everything works now! :+1:

PR #395, opened 2024-02-07
feat(init): add generic error page

d-loose:
Adds an error page to the init and welcome wizards that is shown whenever loading the next page in the wizard fails (e.g. due to grpc errors).
I didn't touch the error handling in ubuntu-bootstrap yet, where the user is taken to the `InstallPage`, which shows the journal output, in case subiquity reports that the installation has failed.

P.S. The UI is only a placeholder that should be replaced once we agree on a design

Ref #363 
UDENG-2204


packages/ubuntu_init/lib/src/init_wizard.dart
```diff
@@ -54,6 +56,18 @@ class InitWizard extends ConsumerWidget {
...
+        Routes.error: WizardRoute(
+          builder: (_) => const ErrorPage(),
+          userData: const WizardRouteData(
+            hasPrevious: false,
+            hasNext: false,
+          ),
+          onNext: (_) async {
+            final window = YaruWindow.of(context);
+            await window.close();
+            return Routes.initial;
```
spydon, line 68: Shouldn't this return null?

packages/ubuntu_init/lib/src/init_wizard.dart
```diff
@@ -105,6 +121,18 @@ class WelcomeWizard extends ConsumerWidget {
...
```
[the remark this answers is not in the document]
d-loose: Yes, `NextWizardButton` only changes its default label from "Next" to "Done" if `hasNext` is false, but still calls `onNext` when the user hits "Done" :mage:

packages/ubuntu_init/lib/src/init_wizard.dart
d-loose: Doesn't really make a difference since we're closing the window - I just copied that part from the final route in the wizard. We could change it to `null` if it's clearer, though

PR #396, opened 2024-02-08
feat: When oem is true in installer, force eula and remove identity

spydon:
When the user sets `oem: true` in the config file the `eula` page should be forced and the `identity` page should be removed from stage 3, if subiquity responds with interactive pages this behavior has no effect.


d-loose (approved):
Thanks, LGTM :+1:

PR #398, opened 2024-02-08
fix: Fetch file images from whitelabel directory

spydon:
All images should be fetched from `/usr/share/desktop-provision/images` so that the users doesn't have to specify the full path (and since we anyways can't support any path.


d-loose (approved):
LGTM :+1:

d-loose:
@spydon something I overlooked here: the file image paths are still expected to be 'absolute' (now within the `/usr/share/desktop-provision/images` directory). So a valid config entry would look like
```yaml
pages:
  locale:
    image: "/myimage.png"
```
That `/` shouldn't be necessary.

spydon:
@d-loose it looks like that shouldn't be needed from this (hard to link lines from the GitHub app):
```
final imagePath = '${ConfigService.whiteLabelDirectory}images/$imageName';
```
But I guess you tried? Some other place that overlooks it?

d-loose:
There's
```dart
...
} else if (isAbsolutPath) {
  loadFutures.add(_loadFile(imageConfig, pageName, context));
...
```
in `preCache`

spydon:
> There's
> ```dart
> ...
> } else if (isAbsolutPath) {
>   loadFutures.add(_loadFile(imageConfig, pageName, context));
> ...
> ```
> in `preCache`

Aaah let's remove it!

PR #399, opened 2024-02-08
fix: Make refresh page hide bottom bar and close button

spydon:
When it is downloading a new version it is now hiding the bottom bar and the close button.


packages/ubuntu_bootstrap/lib/pages/refresh/refresh_page.dart
```diff
@@ -25,16 +25,19 @@ class RefreshPage extends ConsumerWidget with ProvisioningPage {
     return WizardPage(
       title: YaruWindowTitleBar(
         title: Text(l10n.refreshPageTitle),
+        isClosable: !model.state.busy,
```
d-loose, line 28: It might be good to figure out why [this](https://github.com/canonical/ubuntu-desktop-provision/blob/c128f3ee4c1aecbad6b5a16a62743c431558ec6a/packages/ubuntu_bootstrap/lib/installer/installer_wizard.dart#L34) doesn't work and remove it, if we find a working solution

spydon: I missed that one! I don't see why one would want to to that from there thought?
That will rebuild the pretty much the whole widget tree instead of only the parts that need to be rebuilt.

d-loose: It used to be called directly via a listener on the model, but got changed [here](https://github.com/canonical/ubuntu-desktop-installer/pull/2104) during some integration test fixing

d-loose (approved):
LGTM :+1:

PR #401, opened 2024-02-09
feat(whitelabel): read app title from top-level config scope

d-loose:
Replaces `window-title` from `theme` scope with `app-name` in top-level scope.


spydon (approved):
Lgtm!

PR #406, opened 2024-02-09
fix: Only show the `AccessibilityPage` if it is supported

spydon:


packages/ubuntu_provision/lib/src/accessibility/accessibility_page.dart
```diff
@@ -13,7 +13,9 @@ class AccessibilityPage extends ConsumerWidget with ProvisioningPage {
   @override
   Future<bool> load(BuildContext context, WidgetRef ref) {
     final model = ref.read(accessibilityModelProvider);
-    return model.init().then((_) => true);
+    // This will return false when it is not Gnome, since the model wont be able
+    // to initialize with GSettings.
+    return model.init().then((_) => true, onError: (_) => false);
```
d-loose, line 18: Did you try this? Errors are caught and logged in `GnomeAccessibilityService`, so `model.init()` should never throw

spydon: I did not, I pushed it during the meeting 😄

d-loose: Just checked it by using nonexistent dconf paths and it doesn't work. Sorry, I think my `GnomeAccessibilityService` implementation fails a bit too gracefully in that case and just replies with default values. Maybe we could throw a custom exception in the `_tryGet`/`_trySet` methods

spydon: What about [this](https://github.com/canonical/ubuntu-desktop-provision/pull/406/commits/67a91f6c18988159f7a240e5f3aafffd364cf20c), if we don't get any options at all back, then we treat it as non-supported?

d-loose: works for me :+1:

d-loose (approved):

packages/ubuntu_provision/lib/src/accessibility/accessibility_page.dart
spydon: Ah, if there are default values I guess that won't work either...

spydon: Seems to work!

spydon: Ah, haha, your comment didn't pop up until I commented

d-loose (changes requested):
Wait sorry again :D the default values are an issue - I thought you had implemented this logic in `_tryGet`

packages/ubuntu_bootstrap/lib/services/gnome_accessibility_service.dart
```diff
@@ -55,6 +55,10 @@ class GnomeAccessibilityService implements AccessibilityService {
     }
   }
 
+  @override
+  Future<bool> isSupported() =>
+      _a11yInterfaceSettings.list().then((_) => true, onError: (_) => false);
```
d-loose, line 60: In principle we should probably check all gsettings interfaces, but that should do for now :)

d-loose (approved):

PR #412, opened 2024-02-12
A11y service impl

### LINKED ISSUE #311: a11y service impl
Implement the a11y grpc service

matthew-hagemann:
closes #311


provd/protos/accessibility.proto
```diff
@@ -4,205 +4,53 @@ package accessibility;
...
-    bool enabled = 1;
-}
-
-message GetHoverClicksResponse {
-    bool enabled = 1;
-}
+    // Zoom
+    rpc GetDesktopZoon(google.protobuf.Empty) returns (google.protobuf.BoolValue) {}
+    rpc EnableDesktopZoon(google.protobuf.Empty) returns (google.protobuf.Empty) {}
+    rpc DisableDesktopZoon(google.protobuf.Empty) returns (google.protobuf.Empty) {}
```
d-loose:
```suggestion
    rpc GetDesktopZoom(google.protobuf.Empty) returns (google.protobuf.BoolValue) {}
    rpc EnableDesktopZoom(google.protobuf.Empty) returns (google.protobuf.Empty) {}
    rpc DisableDesktopZoom(google.protobuf.Empty) returns (google.protobuf.Empty) {}
```

matthew-hagemann: hehe, okay, i've installed a spellchecked :sweat_smile:

provd/internal/services/accessibility/accessibility.go
```diff
@@ -0,0 +1,121 @@
...
+	if !success {
+		return nil, status.Errorf(codes.Internal, errorFmt, enabled)
+	}
+
+	return &emptypb.Empty{}, nil
+}
+
+func getBooleanSettings(boolSettings booleanGetter, key string) bool {
+	return boolSettings.GetBoolean(key)
+}
```
matthew-hagemann, line 108: There is no lines of code gain using a getter, but I thought it looked more consistent along side the setter where we do make those gains.

provd/internal/services/user/actions
```diff
@@ -0,0 +1,34 @@
+Accounts.FindUserByName(name: mock-user)
```
didrocks, line 1: hum, I think that was committed by mistake?

provd/internal/services/timezone/testdata/TestGetTimezone/golden/error_on_empty_request
```diff
@@ -0,0 +1 @@
+<nil>
```
didrocks, line 1: See, we shouldn’t have this golden file.

didrocks:
@matthew-hagemann are you waiting for me on that one? It doesn’t seem my comments are addressed or acked (and you didn’t request for review).

Just asking, fine if you didn’t have time, just to ensure that you didn’t miss them :)

matthew-hagemann:
> @matthew-hagemann are you waiting for me on that one? It doesn’t seem my comments are addressed or acked (and you didn’t request for review).
> 
> Just asking, fine if you didn’t have time, just to ensure that you didn’t miss them :)

Hi @didrocks! Sorry I didn't ack, and ty for the review! We shifted focus last week to packaging and prepping for demos, so I haven't had a chance to circle back to features yet, but will address these first as soon as I do :slightly_smiling_face:

didrocks:
> Hi @didrocks! Sorry I didn't ack, and ty for the review! We shifted focus last week to packaging and prepping for demos, so I haven't had a chance to circle back to features yet, but will address these first as soon as I do 🙂

No worry, I was just wondering if that was an oversight :) Good luck with the demo!

provd/internal/services/accessibility/accessibility.go
```diff
@@ -0,0 +1,121 @@
...
+	if !isWritable {
+		return nil, status.Errorf(codes.Internal, "failed to connect to location settings")
+	}
+
+	isWritable = s.wmSettings.IsWritable("audible-bell")
+	if !isWritable {
+		return nil, status.Errorf(codes.Internal, "failed to connect to location settings")
+	}
+
+	isWritable = s.keyboardSettings.IsWritable("sticky-keys")
```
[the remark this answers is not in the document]
matthew-hagemann: Thanks! At the moment I don't think I'll add anymore, I just picked one per `GSettings` object as a "ping" test, but will keep this in mind :slightly_smiling_face:

provd/internal/services/accessibility/accessibility_test.go
```diff
@@ -0,0 +1,169 @@
...
+	if g.setDoubleError {
+		return false
+	}
+	g.currentDouble = value
+	return true
+}
+
+func (g gSettingsSubsetMock) GetDouble(key string) float64 {
+	if g.getDoubleError {
+		return 0
```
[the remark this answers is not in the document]
matthew-hagemann: My understanding of the underlying func this calls is that is will just return the default value if somthing goes wrong:
```
func (this0 *Settings) GetDouble(key0 string) float64 {
	var this1 *C.GSettings
	var key1 *C.char
	if this0 != nil {
		this1 = (*C.GSettings)(this0.InheritedFromGSettings())
	}
	key1 = _GoStringToGString(key0)
	defer C.free(unsafe.Pointer(key1))
	ret1 := C.g_settings_get_double(this1, key1)
	var ret2 float64
```
It doesn't return a success bool so we can't tell if there was a failure?

provd/internal/services/accessibility/hearing_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+		// Wants
+		wantTrue bool
+	}{
+		// Success case
+		"Return visual alerts are disabled": {},
+		"Return visual alerts are enabled":  {wantTrue: true},
+
+		// Error cases
+		"Error when can't get state of visual alerts": {getBooleanError: true},
```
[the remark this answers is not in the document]
matthew-hagemann: I have a similar issue to the one I have with `GetBoolean`, there is no success bool returned, only the default value, on error:

```
func (this0 *Settings) GetBoolean(key0 string) bool {
	var this1 *C.GSettings
	var key1 *C.char
	if this0 != nil {
		this1 = (*C.GSettings)(this0.InheritedFromGSettings())
	}
	key1 = _GoStringToGString(key0)
	defer C.free(unsafe.Pointer(key1))
	ret1 := C.g_settings_get_boolean(this1, key1)
	var ret2 bool
```

provd/internal/services/testdata/TestRegisterGRPCServices/golden
```diff
@@ -126,10 +126,10 @@ privacy.PrivacyService:
     metadata: protos/privacy.proto
 timezone.TimezoneService:
     methods:
-        - name: SetTimezone
+        - name: GetTimezone
```
didrocks, line 129: hum, I am puzzled about the order (and the reordering of all those calls), are the golden file reproductible, always printing the same state order everytime?

provd/internal/services/accessibility/accessibility_test.go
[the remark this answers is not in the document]
didrocks: Exactly! That was my point: the test is lying telling that we error out :) So the idea is really to set to some value and check the return value. If there are 2 values, try both :) but don’t tell there is an error we can trigger (because we can from the mock, but not in reality).

Consequently, the mock should not allow to fail either to reflect the real behaviour.

provd/internal/services/accessibility/hearing_test.go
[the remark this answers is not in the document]
didrocks: see previous reply.

didrocks (changes requested):
And here we go! Thanks for all the fixes.

The last 2 remaining comments are:
- check the golden file actions are in the correct order and reproducible (always return the same one so that we don’t start getting flaky tests)
- rework the tests + mock to not make people believe we can get an error on the `Get*` functions.

Good work! I think we are close to merging this one :)

provd/internal/services/testdata/TestRegisterGRPCServices/golden
matthew-hagemann: No, they actually update the order every time I see.

Here are two outputs of running with `--update`, one after the other

```
diff --git a/provd/internal/services/testdata/TestRegisterGRPCServices/golden b/provd/internal/services/testdata/TestRegisterGRPCServices/golden
index 80ed5885..cf3b1da0 100644
--- a/provd/internal/services/testdata/TestRegisterGRPCServices/golden
+++ b/provd/internal/services/testdata/TestRegisterGRPCServices/golden
@@ -1,107 +1,107 @@
 accessibility.AccessibilityService:
     methods:
-        - name: GetScreenKeyboard
+        - name: GetVisualAlerts
           isclientstream: false
           isserverstream: false
-        - name: GetStickyKeys
+        - name: EnableVisualAlerts
           isclientstream: false
           isserverstream: false 
...
```

```
diff --git a/provd/internal/services/testdata/TestRegisterGRPCServices/golden b/provd/internal/services/testdata/TestRegisterGRPCServices/golden
index 80ed5885..bb1c558c 100644
--- a/provd/internal/services/testdata/TestRegisterGRPCServices/golden
+++ b/provd/internal/services/testdata/TestRegisterGRPCServices/golden
@@ -1,107 +1,107 @@
 accessibility.AccessibilityService:
     methods:
-        - name: GetScreenKeyboard
+        - name: GetHighContrast
           isclientstream: false
           isserverstream: false
-        - name: GetStickyKeys
+        - name: GetSlowKeys
           isclientstream: false
           isserverstream: false
...
```

matthew-hagemann:
@didrocks I've updated the `Get*` tests and removed false error cases as well as add a test for when a default double is returned :slightly_smiling_face:

provd/internal/services/testdata/TestRegisterGRPCServices/golden
didrocks: Yeah, so the order is not deterministic. It’s definitively something that has to be fixed/figured out before merging this, or you will go crazy on every update :)

As we calling them in a non deterministic order in those tests? Is it something expected and only what matters is what get called in terms of how many times they are called with each parameters?
If so, we can have some reorder function that we intercept before saving in the golden file.

didrocks:
Excellent! (unsure why there is one of the comment I can’t resolve). So the last thing to figure out is the golden file update to always save the call in a deterministic way. (see my comment on the opened conversation), correct?

didrocks (approved):
Once you removed the printf!

d-loose (approved):
LGTM :+1:

PR #413, opened 2024-02-12
feat(init): migrate to provd privacy service

d-loose:
Replaced the gnome privacy service with a provd-based one.

Ref: #363
UDENG-2204


spydon (approved):
Lgtm!

PR #419, opened 2024-02-13
feat: Use AudioPlayers instead of canberra-gtk-play

### LINKED ISSUE #416: Don't use `canberra-gtk-play` to play sounds
Since `canberra-gtk-play` won't be present on most flavors we shouldn't use this to play sounds, or alternatively only use it when it's present.

There are a few different options:
1. Don't play any sound when `canberra-gtk-play` isn't present
2. Use something that is present on more systems instead
3. Bundle the welcome sound in assets and use AudioPlayers or similar package to play it
3.1 Allow configuration of what welcome sound that should be played

spydon:
This means we also have to include the following packages in the snap:
* liblzma-dev (might already be included)
* libgstreamer1.0-dev
* libgstreamer-plugins-base1.0-dev

And optionally if we want support for more formats later (probably not needed):
* gstreamer1.0-plugins-good
* gstreamer1.0-plugins-bad

Closes:  #416


ogres:
@spydon Is this still valid?

spydon:
> @spydon Is this still valid?

It is, but I think the maintainers will have to decide on whether it is worth changing to another sound dependency, so I'll close this.

PR #420, opened 2024-02-13
feat(init): add a11y page and provd service

d-loose:
Adds a provd-based accessibility service for the wizard. 

Ref: #363
UDENG-2204


packages/ubuntu_init/integration_test/e2e_test.dart
```diff
@@ -29,6 +29,11 @@ void main() {
     await tester.tapNext();
     await tester.pumpAndSettle();
 
+    await tester.testAccessibilityPage();
+    await tester.tapNext();
+    await tester.pumpAndSettle();
+    // TODO: Test accessibility settings
```
d-loose, line 35: Just like for the other gsettings-based services proper assertions are still missing here, as we haven't found a working solution for the CI yet.
However, this test ensures that the corresponding endpoint in provd responds successfully.

packages/ubuntu_init/lib/src/routes.dart
d-loose, line 1: Overlooked this file while rebasing - will remove it, as it's no longer needed

spydon (approved):
Lgtm!

PR #423, opened 2024-02-14
feat: read flavor from config

### LINKED ISSUE #415: Add 'flavor' option to whitelabel.yaml
To improve and simplify support for flavors, we plan to add a `flavor` option to the config file which determines the distro name used in l10n strings as well as the default yaru accent color.

d-loose:
Parses the `flavor` from the config file. If it can't be parsed (i.e. it's not one of budgie, cinnamon, edubuntu, kubuntu, kylin, lubuntu, mate, studio, ubuntu, unity or xubuntu) it uses [`ubuntu_flavor`](https://github.com/canonical/ubuntu-flutter-plugins/tree/main/packages/ubuntu_flavor)'s `detect()`, which tries to determine the flavor from the value of `XDG_CURRENT_DESKTOP`. 'ubuntu' is used as a fallback.

The flavor determines the primary color of the yaru theme as well as the `DISTRO` name used in l10n strings.

Closes #415
UDENG-2300

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -39,7 +39,7 @@ command:
       split_view: ^3.2.1
       stdlibc: ^0.1.4
       synchronized: ^3.1.0
-      ubuntu_flavor: ^0.2.2
+      ubuntu_flavor: ^0.3.0
       ubuntu_localizations: ^0.3.5
       ubuntu_logger: ^0.1.1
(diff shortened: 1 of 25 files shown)


spydon (approved):
Lgtm!
(We had `displayName` eeeeverywhere on my last job :sweat_smile:)
tenor

PR #426, opened 2024-02-15
fix(whitelabel): don't require 'absolute' paths for images

d-loose:
Currently the page image configuration only works when specified like this
```yaml
pages:
  locale:
    image: "/foo.png"
```

This PR removes the need for the leading `/` and improves the test.


spydon (approved):
Lgtm, my oversight

PR #430, opened 2024-02-16
Pro service protos

matthew-hagemann:
Updated pro protos to account for the magic attach flow

Q: Do we need a standalone pro attach rpc or are we only supporting magic attach?
A: Yes

Screenshot from 2024-02-15 15-23-13


d-loose (approved):
LGTM :+1:

PR #433, opened 2024-02-16
fix(provd): bus is up if unix:path exists

nteodosio:
I had a problem in my machine when copying these testutils. I hit

> system bus mock is not running. If that's intended, manually connect to the system bus instead of using this function

after calling both 
```
StartLocalSystemBus()
GetSystemBusConnection()
```

Turned out my system would create a unix:path instead of a unix:abstract socket.

Either of them existing means that the private Dbus is running and so I propose checking for both.

### DIFF
--- a/provd/internal/testutils/systembus.go
+++ b/provd/internal/testutils/systembus.go
@@ -142,5 +142,6 @@ func GetSystemBusConnection() (*dbus.Conn, error) {
 
 // isRunning checks if the system bus mock is running.
 func isRunning() bool {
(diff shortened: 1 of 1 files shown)


d-loose (approved):
Awesome, I ran into the same issue as well!

nteodosio:
Great! Apparently it will take a time until I find out how to make Git use the SSH key I use in Github. (:

matthew-hagemann (approved):
Thank you @nteodosio!

PR #434, opened 2024-02-16
feat: remove flavor detection

d-loose:
In order to avoid any confusion, we've decided to simplify the configuration for flavors. This PR removes the automatic detection mechanism based on `XDG_CURRENT_DESKTOP`, so that the flavor needs to be set explicitly via the `flavor` key in `/usr/share/desktop-provision/whitelabel.yaml`.


spydon (approved):
Lgtm

PR #437, opened 2024-02-16
fix: Don't crash if `canberra-gtk-play` doesn't exist

spydon:
This is a quick workaround so that we don't crash on systems without `canberra-gtk-play`, until #419 is done.

### DIFF
--- a/packages/ubuntu_provision/lib/src/services/sound_service.dart
+++ b/packages/ubuntu_provision/lib/src/services/sound_service.dart
@@ -9,7 +9,7 @@ class SoundService {
   Future<void> play(String id) async {
     try {
       await Process.run('canberra-gtk-play', ['--id=$id']);
-    } on ProcessException catch (e) {
+    } on ProcessException catch (_) {
       _log.error(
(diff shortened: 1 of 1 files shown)


packages/ubuntu_provision/lib/src/services/sound_service.dart
```diff
@@ -1,7 +1,18 @@
...
+
+/// A service for playing sounds.
 class SoundService {
   Future<void> play(String id) async {
-    await Process.run('canberra-gtk-play', ['--id=$id']);
+    try {
+      await Process.run('canberra-gtk-play', ['--id=$id']);
+    } on ProcessException catch (e) {
+      _log.error(
+        'Error playing sound with id: $id (most likely missing canberra-gtk-play)',
```
d-loose, line 14: You need to either log `e` as well, or remove the variable

d-loose (approved):
Besides the small comment, LGTM :+1:

packages/ubuntu_provision/lib/src/services/sound_service.dart
spydon: Ah ofc!

PR #241, opened 2023-11-30
feat(provd): adding debian file & systemd files

matthew-hagemann:
- Debian files for the provisioning backend

Closes #479 
Closes #480

PR #442, opened 2024-02-19
test: add end-to-end test for ubuntu-init + provd

### LINKED ISSUE #411: end-to-end test for ubuntu init + provd
In order to ensure that the communication between ubuntu init and provd works as intended, we should set up an end-to-end test.
As a starting point, we can run provd as root and use a flutter integration test to drive the UI.

d-loose:
Current CI setup:
* Manually run `provd` as `root`
* Allow everyone `rw` access to `/run/provd/provd.sock`

An issue with that setup is, that `provd` can't change any gsettings options:
```
(process:16409): dconf-WARNING **: 10:58:40.715: failed to commit changes to dconf: Cannot autolaunch D-Bus without X11 $DISPLAY
```
I've left todo notes to add tests for the privacy service and the gnome keyboard input sources that rely on those, once we have a solution.

Once we figure the details on how to run `provd` as the `gnome-initial-setup` and use systemd socket activation, we should do that in the e2e test as well.

For reference, here's what I've tried so far (using the systemd files from #241):
```yaml
...
- name: Setup polkit rules
  run: |
    sed "s/gnome-initial-setup/$USER/g" 20-gnome-initial-setup.rules | sudo tee 20-$USER.rules
    sudo sed "s/gnome-initial-setup/$USER/g" /var/lib/polkit-1/localauthority/10-vendor.d/gnome-initial-setup.pkla | sudo tee /var/lib/polkit-1/localauthority/10-vendor.d/$USER.pkla
  working-directory: /usr/share/polkit-1/rules.d
- name: Run provd
  run: |
    sudo cp provd/provd /usr/sbin/
    sudo cp provd/systemd/provd.service /lib/systemd/system/
    sudo cp provd/systemd/provd.socket /lib/systemd/system/
    sudo sed -i "s/gnome-initial-setup/$USER/" /lib/systemd/system/provd.socket
    sudo systemctl daemon-reload
    sudo systemctl start provd.socket

...
```
I kept running into dbus permission issues, though.

Close #411 
UDENG-2286

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
       run: |
         set -ux
         set +e
-        sed 's|test.sock|provd/provd.sock|' provd/example.config.yaml  > provd/provd.yaml
+        mv provd/example.config.yaml provd/provd.yaml
         sudo mkdir -p /run/provd
         sudo chown $USER /run/provd
         sudo provd/provd &
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,42 @@ jobs:
     - uses: bluefireteam/melos-action@v2
     - run: melos format
 
+  init-e2e:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+    - uses: asdf-vm/actions/install@v3
+    - uses: bluefireteam/melos-action@v2
+    - name: Install dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y clang cmake curl libgtk-3-dev ninja-build pkg-config unzip
+        sudo apt install -y gnome-initial-setup network-manager
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: "./provd/go.mod"
+    - name: Build provd
+      run: go build ./cmd/provd
+      working-directory: provd
+    - name: Run provd
+      run: |
+        set -ux
+        set +e
+        sed 's|test.sock|provd/provd.sock|' provd/example.config.yaml  > provd/provd.yaml
+        sudo mkdir -p /run/provd
+        sudo chown $USER /run/provd
+        sudo provd/provd &
+        until [ -S /run/provd/provd.sock ]; do
+          sleep 1
+        done
+        sudo chmod go+rw /run/provd/provd.sock
+
+    - run: xvfb-run -a -s '-screen 0 1024x768x24 +extension GLX' flutter test integration_test/e2e_test.dart
+      working-directory: packages/ubuntu_init
+
   init:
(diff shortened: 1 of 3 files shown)


spydon (approved):

.github/workflows/ci.yml
```diff
@@ -63,6 +63,42 @@ jobs:
...
+      with:
+        go-version-file: "./provd/go.mod"
+    - name: Build provd
+      run: go build ./cmd/provd
+      working-directory: provd
+    - name: Run provd
+      run: |
+        set -ux
+        set +e
+        sed 's|test.sock|provd/provd.sock|' provd/example.config.yaml  > provd/provd.yaml
```
matthew-hagemann, line 90: Maybe it is worth updating  `example.comfig.yaml ` / creating a new one specifically for the e2e tests? Will make this more robust in the event someone accidentally changes the example config and not update the e2e test?

packages/ubuntu_init/integration_test/e2e_test.dart
```diff
@@ -0,0 +1,113 @@
...
+    await tester.pumpAndSettle();
+
+    await expectCommand('localectl', [], [
+      contains('LANG=de_DE.UTF-8'),
+      contains('X11 Layout: gb'),
+    ]);
+
+    // TODO: Test gsettings input-sources
+
+    await tester.testNetworkPage(mode: ConnectMode.none);
```
matthew-hagemann, line 43: Only test I can think of would be maybe to check the interface shown in the network page lines up with netplan? But otherwise I imagine this is difficult to test, especially since most the UI and interactions exist for wifi setups :sweat_smile:

matthew-hagemann (approved):
Awesome!

.github/workflows/ci.yml
d-loose: Updating the example config (and maybe renaming it to 'default') sounds like a good idea.

matthew-hagemann: Sounds perfect :+1:

packages/ubuntu_init/integration_test/e2e_test.dart
d-loose: It's a bit cumbersome to deal with the network setup of a github runner. Also since networkmanager and the respective libraries are well-tested elsewhere, I don't think we need to test it extensively here. It's about testing the communication between the UI and provd after all.

.github/workflows/ci.yml
d-loose: Hm, since provd expects a `provd.yaml` renaming 'example' to 'default' is probably misleading, so I'd just update the path

packages/ubuntu_init/integration_test/e2e_test.dart
matthew-hagemann: Fair enough :smile:

PR #443, opened 2024-02-19
ci: Automate rebase to support multiple snaps

tim-hm:
Launchpad lacks a pre-build hook for snap builders; so without changes
we cannot build multiple snaps from one branch. Using a GitHub action
we could reconfigure the repo and then invoke a remote build. But we'd
lose Launchpad's audit trail which is needed for a seeded snap.
Alternatively, we could continue using branches-per-snap [1][2] but this
means silo'ing ci artefacts to branches.

This commit triggers on workflow `main` changes to resolve this. The
workflow:

1. rebases `ci/*` branches on `main`.
2. repositions artefacts for the Launchpad builder jobs.
3. amends the HEAD commit to avoid piling up automated commits.

Note: The workflow does not sign step 3's amended.

References:

[1] https://launchpad.net/~ubuntu-desktop/ubuntu-desktop-provision/+snap/ubuntu-desktop-bootstrap
[2] https://launchpad.net/~ubuntu-desktop/ubuntu-desktop-provision/+snap/ubuntu-desktop-init


.github/workflows/rebase-ci-bootstrap.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - main
+
+jobs:
+  rebase:
+    runs-on: ubuntu-22.04
+    steps:
+
+    - uses: actions/checkout@v4
+      with:
+        ref: ci/bootstrap
```
spydon, line 15: This could be run as a matrix job instead to avoid duplication, if one would like to.
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

spydon (approved):
Lgtm, just one small improvement that could be made

d-loose (approved):
I also had a small suggestion in line with Lukas' idea.
Besides that, LGTM :+1:

.github/workflows/rebase-ci-bootstrap.yaml
tim-hm: I did consider this ... but then had to consciously resist the urgh to optimise :) Wdty about going with the simpler approach for now and then revisiting if/when we expand this out to two cases?

.github/workflows/rebase-ci-bootstrap.yaml
```diff
@@ -0,0 +1,35 @@
...
+        git config --global user.name "GitHub Actions"
+
+        git fetch origin main:main
+        git rebase -X theirs main
+
+        mv snap snap.tmp
+        mv snap.tmp/bootstrap snap/
+        rm -rf snap.tmp
+
+        ./scripts/subiquity_integration
```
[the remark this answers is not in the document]
tim-hm: I agree with this. But looking at ci.yaml and considering lakas' comments above, I actually think we can do this as part of a bigger ci refactor. I'll create an issue with my proposal and get this merged as is.

PR #451, opened 2024-02-21
fix: Avoid double loading

spydon:
This also removes the need for `context` in the `PageImagesProvider` so that it can load from a `FutureProvider`.

We should later probably move more things into the `loadingProvider`.


d-loose (approved):
Awesome, thanks! :+1:

PR #452, opened 2024-02-21
Refine snap ci workflow and improve resiliency

tim-hm:


.github/workflows/pr.yaml
```diff
@@ -23,3 +23,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@v5
+
+  consistency-check:
```
d-loose, line 27: ci.yaml might be a better fit for this

.github/workflows/rebase-for-snap-cicd.yaml
```diff
@@ -6,13 +6,15 @@ on:
       - main
 
 jobs:
-  rebase:
+  rebase:
```
d-loose, line 9:
```suggestion
  rebase:
```

ci/snap/bootstrap/snapcraft.yaml
```diff
@@ -8,8 +8,11 @@ grade: stable
 confinement: classic
 base: core22
 issues: https://bugs.launchpad.net/ubuntu-desktop-provision/+filebug
-contact: https://bugs.launchpad.net/ubuntu-desktop-provision/+filebug
-
+contact: https://github.com/canonical/ubuntu-desktop-provision/issues
```
d-loose, line 11: Didn't we want to use the launchpad bug tracker as an outside facing point of contact?

ci/snap-build-setup.sh
```diff
@@ -0,0 +1,16 @@
...
+if [[ $snap_name != "init" && $snap_name != "bootstrap" ]]; then
+    echo "Expected argument to be 'init' or 'bootstrap'"
+    exit 1
+fi
+
+snapcraft_artifacts_path="ci/snap/${snap_name}"
+
+cp -r "${snapcraft_artifacts_path}" ./snap
+
+# inside subiquity client
```
spydon, line 16: Remove comment?

.github/workflows/rebase-for-snap-cicd.yaml
```diff
@@ -21,16 +23,10 @@ jobs:
...
-        rm -rf snap.tmp
+        git reset --hard main
 
-        ./scripts/subiquity_integration
+        ./ci/snap-build-setup.sh ${{ matrix.input }}
         
-        COMMIT_MSG="ci(bootstrap): prep for launchpad snap builder"
         git add -A
-        git commit --amend -m "$COMMIT_MSG" --author="GitHub Actions <github-actions@github.com>"
+        git commit --amend --no-edit # keep original author but set committer to github-actions@github.com
```
spydon, line 31: If we're changing the SHA anyways, isn't it better to just create a new commit?

ci/snap/bootstrap/snapcraft.yaml
```diff
@@ -155,9 +163,9 @@ parts:
 
   ubuntu-bootstrap:
     after: [flutter-git]
-    source: .
-    source-type: git
     plugin: nil
+    source: https://github.com/canonical/ubuntu-desktop-provision.git
```
d-loose, line 167: Hm, I've only ever seen `source: .` in snapcraft.yamls - I'd double check with someone more experienced if this is a good thing to do. Personally I'd find it rather confusing since I'd expect `snapcraft` to build from my local repo

ci/snap/init/snapcraft.yaml
```diff
@@ -6,7 +6,7 @@ description: |
 grade: stable
 confinement: strict
 base: core22
-issues: https://github.com/canonical/ubuntu-desktop-provision/issues
+issues: https://bugs.launchpad.net/ubuntu-desktop-provision/+filebug
 contact: https://github.com/canonical/ubuntu-desktop-provision/issues
```
d-loose, line 10: See above, shouldn't we change this to launchpad as well?

.github/workflows/pr.yaml
tim-hm: Lol yes of course it is 😂

.github/workflows/rebase-for-snap-cicd.yaml
tim-hm: Resolved in latest formatting commit

ci/snap-build-setup.sh
tim-hm: Nice catch

.github/workflows/rebase-for-snap-cicd.yaml
tim-hm: We touched on this during standup and both approaches have flaws. If you add a commit then the message show in launchpad isn't helpful (ie all you see is something like `ci: prep for launchpad snap build`). If you modify the HEAD then it's not signed but at least the commit message matches what is in the git history.

I'm happy either way, so let me know if you have a strong opinion either way!

ci/snap/bootstrap/snapcraft.yaml
tim-hm: I agree for issues but I wondered if for contact we sent folks to github or even provide an email address? Ie if someone is on snap store and clicks contact us sending them to launchpad might not quiet be what we want. 

WDTY?

ci/snap/bootstrap/snapcraft.yaml
tim-hm: There isn't a better/worse approach both as `source: .` and `source: <url>` are supported references; however they behave differently and so we need to consider that

With `source: .` + local dev, a whack of artefacts make their way into the container and that leads to non-deterministic builds.  We might want this because we iterate on local changes. But critically, we know how to clear artefacts to support local iterations. On my first pass of this PR, I build a script to do this something along the lines of:

```bash
$ flutter clean # needed because the cache has some paths which include $USER
$ snapcraft clean # potentially needed depending on how previous builds went / what changes you made
$ rm -rf packages/subiquity_client/subiquity/.subiquity #  snapcraft fails if artefact is left behind from a local run
$ rm -rf snap # might be needed if the last time you ran it was against the other snap artefact, but also this folder is required for the snapcraft command to run against the right container
```

That's a lot of nuance and I don't think it's worth the hassle, especially because once you know all the nuances its straightforward to run local builds. So, my thinking turned towards the public. Ie someone who might clone the repo, modify locally, and then try and build the snap using `source: .`. In this instance they're going to quickly run into issues and then they'll ask for help or spend a chunk of time tracing through everything to figure out whats going on. I think this group is who we should target with the repos default behaviour because its straightforward to modify the snapcraft.yaml to pull in local changes.

If you'd prefer `source: .` happy to go with that too!

ci/snap/bootstrap/snapcraft.yaml
d-loose: What does "contact" mean in this case anyways? I'd just like to avoid having bug reports spread across github and launchpad again, since it's chaotic enough already with old issues from ubuntu-desktop-installer.

ci/snap/bootstrap/snapcraft.yaml
d-loose: I also noticed we don't have any workflows that build the snaps in the github CI now (they used to run on the bootstrap/init-snap branches). If we want to add them back for QA testing on PRs, `source: https://github...` wouldn't work as intended.

.github/workflows/rebase-for-snap-cicd.yaml
d-loose: In any case, we should introduce meaningful version numbers for the snaps then, as the associated commit hash will never be part of `main`.

ci/snap/bootstrap/snapcraft.yaml
tim-hm: That's a good pt 🤔  as we defo want to improve our pipeline testing. Okay how about this:

1. Revert to `source: .`
2. I'll update `CONTRIBUTING.md` to explain the nuance around building locally and provide some guidance on how/where things can go wrong

.github/workflows/rebase-for-snap-cicd.yaml
tim-hm: Ugh so many issues all because we can't simply do `snapcraft -f path-to-thing.yaml`! Lets discuss at standup. I'd like to maintain the sha over version nos.

spydon: We could just add the last commit message to the new commit. i.e.
```
ci(release): <Last commit message>
```

ci/snap/bootstrap/snapcraft.yaml
d-loose: Sounds good :+1:

ci/snap/bootstrap/snapcraft.yaml
tim-hm: Did some investigating. The BLUF is that I think we should drop both contact and issue from the snapcraft.yaml 
 because they're misleading. The fields themselves are described [here](https://snapcraft.io/docs/snapcraft-yaml-reference). Okay the nitty gritty:

Afaict `contact` is displayed in three places: `snap info <name>`, the App Center, and snapstore.io. The values set in the snapstore.io publisher page overrides the snapcraft.yaml. 

On `issues` I couldn't find the `snapcraft.yaml` value anywhere. If you set it in the snapstore.io publisher page then the only place I could find it was on the snapstore.io entry under "Report a bug".

ci/snap/init/snapcraft.yaml
tim-hm: See discussion above.

.github/workflows/rebase-for-snap-cicd.yaml
tim-hm: Correct me if I'm off, but is the aim to be able to trace the build artefact back to a specific commit sha on main?

If so, then piling release commits on `ci/name` might work at the start, but over time the branch's history would become long and problematic. For example, after 50 releases you'd have to go quite far back in the launchpad history to find the HEAD of main -- although it would still be possible! If we amend to avoid stacking commits, the commit in `ci/<name>` would be overridden on the next invocation and so tracing back to a specific sha would only be a matter of looking at HEAD^1 in launchpads history.

I think the solution we actually want is to combine a few bits here:

- I like your updated commit message so will adopt that
- I think we can use [external-metadata](https://snapcraft.io/docs/using-external-metadata) to get info from HEAD^1 to properly version the artefact.

WDTY?

spydon: > If so, then piling release commits on ci/name might work at the start

Yeah, that doesn't sound like a good solution.

> I like your updated commit message so will adopt that

In a new commit right? Maybe we should add the SHA of the previous commit in the commit message too, then we'd be able to track it "properly", or maybe not worth it since it'll always be HEAD~1?

> I think we can use [external-metadata](https://snapcraft.io/docs/using-external-metadata) to get info from HEAD^1 to properly version the artefact.

Sounds good, we should probably start running `melos version` as part of this workflow too and get the version from there.

d-loose: A dedicated release commit message would be great :+1:
Would it be terribly misleading if we used `snapcraftctl set-version <hash of HEAD~>`?
I'm also not opposed to the `melos version` idea :)

tim-hm: Yep agree with all of this. I've created https://github.com/canonical/ubuntu-desktop-provision/issues/456 to get that work done if you're happy to punt that challenge for the moment?

spydon (approved):

d-loose (approved):

PR #453, opened 2024-02-21
In core desktop only allow to erase the disk

sergio-costas:
Core desktop is installed by copying a disk image directly to the hard disk, so other options like partitioning or installing along the current operating systems aren't supported.

This patch removes all those options when installing core desktop.

### DIFF
--- a/packages/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
@@ -136,7 +136,8 @@ class StorageModel extends SafeChangeNotifier {
   }
 
   /// Whether any advanced features are available.
-  bool get hasAdvancedFeatures => (hasLvm || hasZfs || hasTpm) && !isCoreDesktop;
+  bool get hasAdvancedFeatures =>
+      (hasLvm || hasZfs || hasTpm) && !isCoreDesktop;
 
   ProvisioningMode? _provisioningMode;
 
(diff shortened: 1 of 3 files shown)


d-loose:
Hm, is this really something we want/need to maintain in the UI, or is there a chance subiquity could handle this by sending the appropriate responses for core desktop?

sergio-costas:
@d-loose I'm not sure... if subiquity knows that we are installing a disk image and that the only possibility is to erase the disk and nothing else, then it would be possible...

sergio-costas:
Ok, I think that it could be done by detecting "dd-*" types.

sergio-costas:
After checking subiquity, I think that there are two ways. The simplest one is to keep this patch, but using `hasDd` instead of checking if it is a Core Desktop installer. The other would be to modify subiquity here:

imagen

because being a `dd-*` type, it should disable GuidedCapability.{LVM, LVM_LUKS, ZFS, CORE_BOOT_ENCRYPTED, CORE_BOOT_PREFER_ENCRYPTED}, and empty the manual targets, the resize targets and the usegap targets.

sergio-costas:
@d-loose It seems that there is some problem with the dependencies... 

```
2024-02-23T16:40:03.3661844Z ##[group]Install tools and dependencies
2024-02-23T16:40:03.5590597Z go: github.com/golangci/golangci-lint@v1.55.2: module github.com/golangci/golangci-lint@v1.55.2 found, but does not contain package github.com/golangci/golangci-lint
2024-02-23T16:40:03.5615509Z ##[error]Process completed with exit code 1.
```

d-loose:
> @d-loose It seems that there is some problem with the dependencies...

Those Go dependency issues should be resolved now - could you rebase this onto main, please?

sergio-costas:
Rebased.

d-loose (approved):
LGTM :+1:

PR #455, opened 2024-02-22
feat(provd): add GDM service

d-loose:
Since the necessary snapd interfaces that would allow launching the user's desktop session via GDM during stage 4 don't exist yet, we need provd to take care of that too.
The implementation follows [gnome-initial-setup](https://gitlab.gnome.org/GNOME/gnome-initial-setup/-/blob/daa154171abe2bd6dc0f42f7e37a2d6622050451/gnome-initial-setup/pages/summary/gis-summary-page.c) and would replace our current [dart implementation](https://github.com/canonical/ubuntu-desktop-provision/blob/af73d43395cbc6e89b4ca7f67d69b171282e2559/packages/ubuntu_init/lib/src/services/gdm_service.dart) (I'd add the respective dart client for the provd service in a separate PR).

P.S. The end-to-end test is not prepared for this yet, since it can't access GDM over DBus. 

Ref #363
UDENG-2204


provd/internal/testutils/gdm_mock.go
```diff
@@ -0,0 +1,222 @@
...
+		return nil, fmt.Errorf("failed to acquire account name on local system bus: %w", err)
+	}
+	if reply != dbus.RequestNameReplyPrimaryOwner {
+		return nil, fmt.Errorf("failed to acquire account name on local system bus: name already taken")
+	}
+
+	return &privateBus{conn, tempDir, address, cleanup}, nil
+}
+
+func startPrivateDbusServer() (string, string, func()) {
```
d-loose, line 158: I've mostly re-used the code from `systembus.go` for this function, but kept it separate for now. It might make sense generalize `systembus.go` and move the private bus stuff there too at some point.

matthew-hagemann: Fair enough, but agree its fine to have a little repetition for now :slightly_smiling_face:

provd/internal/services/gdm/gdm.go
```diff
@@ -0,0 +1,148 @@
...
+				s.userVerifier.Call(consts.DbusGdmPrefix+".UserVerifier.AnswerQuery", 0, msg, req.Password)
+			// SessionOpened signal
+			case consts.DbusGdmPrefix + ".Greeter.SessionOpened":
+				slog.Debug(fmt.Sprintf("SessionOpened signal received: %#v", signal))
+				msg, ok := signal.Body[0].(string)
+				if !ok {
+					return nil, status.Errorf(codes.Internal, "failed to get message from SessionOpened signal")
+				}
+				// Tell GDM to start the session and return
+				s.greeter.Call(consts.DbusGdmPrefix+".Greeter.StartSessionWhenReady", 0, msg, true)
```
matthew-hagemann, line 134: 1: What you could do with your calls to `StastSessionWhenReady` and `AnswerQuery` is look for returned from dbus and catch the cases where your side effects don`t succeed.

provd/internal/services/gdm/gdm_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestLaunchDesktopSession(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		username  string
+		password  string
+		cancelReq bool
+
+		wantErr bool
```
matthew-hagemann, line 65: 2: ...You could then have flags for cases where your side effects fail that you pass to the mocks

provd/internal/testutils/gdm_mock.go
```diff
@@ -0,0 +1,222 @@
...
+	if g.username != "myuser" || answer != "mypassword" {
+		_ = emitSignal(g.conn, dbus.ObjectPath("/org/gnome/DisplayManager/Session"), string(s), consts.DbusGdmPrefix+".UserVerifier.Problem", "gdm-password", "Invalid username or password")
+	}
+	_ = emitSignal(g.conn, dbus.ObjectPath("/org/gnome/DisplayManager/Session"), string(s), consts.DbusGdmPrefix+".Greeter.SessionOpened", "gdm-password")
+	return nil
+}
+
+func (g *gdmSessionBus) StartSessionWhenReady(serviceName string, shouldStartSession bool) *dbus.Error {
+	return nil
+}
```
matthew-hagemann, line 89: 3: ...and in your mocks return errors when those flags are set to simulate those cases

matthew-hagemann (commented):
Really nice work! I left a few comments re: testing side effects, but otherwise looks good! :smile:

provd/internal/services/gdm/gdm.go
d-loose: Thanks, very good point - will do!

provd/internal/services/gdm/gdm.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// Listen for signals on the session bus
+	signals := make(chan *dbus.Signal)
+	defer close(signals)
+	s.privateConn.Signal(signals)
+	defer s.privateConn.RemoveSignal(signals)
+
+	// Begin user verification
+	s.userVerifier.Call(consts.DbusGdmPrefix+".UserVerifier.BeginVerificationForUser", 0, "gdm-password", req.Username)
```
d-loose, line 104: @matthew-hagemann I suppose I should catch the error here too and add a test case as well?

matthew-hagemann: Ah, yes please 🙂

d-loose:
I've added error handling for the cases mentioned above.

I'm wondering if we should make the entire GDM service optional in provd. Currently provd fails to start if calling `org.gnome.DisplayManager.Manager.OpenSession` fails (which is typically the case if you're testing it in a normal user session: `GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Can only be called before user is logged in`).

matthew-hagemann:
> I've added error handling for the cases mentioned above.
> 
> I'm wondering if we should make the entire GDM service optional in provd. Currently provd fails to start if calling `org.gnome.DisplayManager.Manager.OpenSession` fails (which is typically the case if you're testing it in a normal user session: `GDBus.Error:org.freedesktop.DBus.Error.AccessDenied: Can only be called before user is logged in`).

Are those the permission issues you were talking about? :thinking: that is a tricky one...

d-loose:
No that's not a permission issue - you simply can't login through GDM if you're already logged in :). It's different in the gnome-initial-setup session, from where you can launch another user's desktop session.

d-loose:
Would it be worth adding an endpoint that simply restarts GDM, just so that we have a working fallback?

PR #458, opened 2024-02-23
fix(init): ensure init model is initialized

d-loose:
Reverts a small change from #451. `onLoad` is never called on the first route in the wizard, so we need to rely on `onReplace` to initialize the model and precache the images


spydon (approved):
Ah, I forgot to change that back, thanks for fixing it!

PR #462, opened 2024-02-27
feat(init): update snap interfaces

d-loose:
Removes the outdated plugs and replaces them with:
- `network` and `network-manager` to access the network and configure network-manager
- `system-files`-based plugs that allow access to various locations in the host file system:
  - `/usr/share/desktop-provision` (r) - configuration directory for whitelabelling
  - `/run/provd/provd.sock` (rw) - communication with provd
  - `/var/log/installer` (w) - logging

What's left to figure out is how to access GDM via DBus. There's no interface that lets us do this using the current dart implementation of the GDM service in ubuntu-init. #455 adds that capability to provd, but we need to figure out how provd can access GDM if it runs as a socket-activated systemd service.

### DIFF
--- a/ci/snap/init/snapcraft.yaml
+++ b/ci/snap/init/snapcraft.yaml
@@ -19,9 +19,11 @@ apps:
     environment:
       LOG_LEVEL: debug
     plugs:
-      - hardware-observe
-      - log-observe
-      - system-observe
+      - usr-share-desktop-provision
+      - run-provd-socket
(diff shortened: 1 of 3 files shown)


matthew-hagemann (approved):
LGTM!

> but we need to figure out how provd can access GDM if it runs as a socket-activated systemd service.

I'll create an issue to look into this to see if I can hack something out on the provd side

PR #464, opened 2024-02-28
fix(init): patch /tmp/ path for gdm service in snap

d-loose:
Until we have sorted out the permission issues for #455, we need to rely on the existing dart implementation of the GDM service to launch the desktop session. Since GDM exposes its private socket in the host's `/tmp/` directory, we need yet another `system-files` plug to access that.

### DIFF
--- a/ci/snap/init/snapcraft.yaml
+++ b/ci/snap/init/snapcraft.yaml
@@ -23,6 +23,7 @@ apps:
       - run-provd-socket
       - network
       - network-manager
+      - tmp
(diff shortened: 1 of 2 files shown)


spydon (approved):
Soon enough we'll have an interface towards root ;D

matthew-hagemann (approved):

d-loose:
I hope none of these things are permanent. Eventually we'll need to create a dedicated snapd interface that gives us permission to all the necessary files and directories.

PR #468, opened 2024-02-28
feat: Slides as "HTML"

spydon:
This converts the slide system to use "HTML" instead, within quotes due to that it is nowhere near the full HTML spec, and has lots of quirks, but it should still work for basic slides.
It first looks in `/usr/share/desktop-provision/slides/` for slides, and if there are none there it takes the slides from assets.


packages/ubuntu_bootstrap/lib/slides/slides_provider.dart
```diff
@@ -0,0 +1,140 @@
...
+        loadFutures.add(slideFuture);
+      }
+    }
+
+    slides.addAll((await Future.wait(loadFutures)).whereType());
+  }
+
+  Future<SlideHtml?> _loadSlide(int index, {bool fromAssets = false}) async {
+    const defaultLocale = 'en_US';
+    final locale = Intl.defaultLocale ?? await findSystemLocale();
```
d-loose, line 54: On the welcome page and in the storage dialogs we use `Localizations.localeOf(context)`. Is there a way we could do the same for the slides, or did you deliberately decide against that?

spydon: We don't have the context in here, it would be the same result right?

d-loose: Ah, probably yes :+1:

d-loose (approved):
LGTM :+1:

JesseRiemens:
> Had to pin dart_style to 2.3.4 due to [dart-lang/dart_style#1413](https://github.com/dart-lang/dart_style/issues/1413)

This issue is fixed now. Should this be reverted?

spydon:
> > Had to pin dart_style to 2.3.4 due to [dart-lang/dart_style#1413](https://github.com/dart-lang/dart_style/issues/1413)
> 
> This issue is fixed now. Should this be reverted?

It's fixed [here](https://github.com/canonical/ubuntu-desktop-provision/pull/471/commits/ceba2a7a9f3c97c30a15bb414efe64427c9832ab) already, just waiting for merge. :)

PR #469, opened 2024-02-29
feat(refresh): quit installer instead of restarting

d-loose:
The refresh mechanism currently fails to launch the updated version of the UI once the ubuntu-desktop-bootstrap snap has been refreshed by subiquity, see https://github.com/canonical/ubuntu-desktop-installer/issues/2377. This is because calling `execv(Platform.resolvedExecutable, [])` from within the snap relaunches the old revision instead of the updated one.
As a simple fix, we can just quit the installer and ask the user to relaunch it. I haven't figured out how to properly test this yet, but I don't see why this approach shouldn't work. In any case, merging this PR cannot make it worse than it currently is :)

### DIFF
--- a/packages/ubuntu_bootstrap/lib/l10n/ubuntu_bootstrap_en.arb
+++ b/packages/ubuntu_bootstrap/lib/l10n/ubuntu_bootstrap_en.arb
@@ -722,7 +722,7 @@
       "snap": {}
     }
   },
-  "refreshRestart": "Please restart the installer.",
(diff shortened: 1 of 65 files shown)


spydon (approved):
Lgtm!
Any reason it is named `onQuit` instead of `onClose` (or maybe `onExit`) for example?

d-loose:
> Lgtm! Any reason it is named `onQuit` instead of `onClose` (or maybe `onExit`) for example?

That was driven by `quitLabel` from the ubuntu localizations.

PR #470, opened 2024-02-29
Update README.md

eeickmeyer:
Apparently this doesn't work unless it's the American spelling. Tried this using `flavour` vs `flavor` and `flavor` won every time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ mode: standard | oem | core-desktop
 # - studio
 # - unity
 # - xubuntu
-flavour: <name>
+flavor: <name>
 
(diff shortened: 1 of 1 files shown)


spydon (approved):
Haha aah, this was mixed up in the other document previously too.

eeickmeyer:
I considered calling the branch `provision-american-revolution` or `provision-american-rebellion`. :joy:

eeickmeyer:
Let me see what I can do. I was only just added on Launchpad to the CLA so that might fix it, so let's see what happens when I fix the email.

spydon:
Super, thanks!

PR #471, opened 2024-03-01
chore: Update to Flutter 3.19.2

spydon:
The biggest change is that we had to migrate to use the `HardwareKeyboard` API.
https://docs.flutter.dev/release/breaking-changes/key-event-migration

Autofocus is removed from the spinbox, since it throws assertions when the number goes down, not a big loss UX-wise.


d-loose (approved):
There's still one unused import - otherwise LGTM :+1:

PR #472, opened 2024-03-01
fix: Swap screenshots to lightmode

spydon:
This swaps some of the images that were added in dark mode to light mode, at a later point we should support dark mode for slides.


d-loose (approved):
Sorry, somehow missed this one - LGTM :+1:

PR #476, opened 2024-03-04
feat: Installation type page design

spydon:
Just the design of the installation type screen, no logic connected to it yet.

image


spydon:
Moving this to draft since it's not the correct screen in the new Figma flow

PR #482, opened 2024-03-05
fix(keyboard_page_test): scroll to item before tapping it

d-loose:
@spydon the keyboard page still looks fine to me on small layouts, although a bit more scrolling is required. Feel free to propose a better solution tomorrow, if you have one :)


spydon (approved):
Thanks! Yeah, I think it works too

PR #484, opened 2024-03-05
Autoinstall page

### LINKED ISSUE #483: Add autoinstall page
We need to add a page that enables users to provide a URL to an autoinstall.yaml during the installation. We should then download the file, verify that it contains valid YAML, copy it to `/` on the live system, and finally restart both subiquity and the UI.

d-loose:
UI Proposal:

[Screencast from 2024-03-05 17-30-12.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/113362648/341f564c-3deb-4606-8644-c6908cdda321)

As discussed, for 24.04 I'd propose showing this page as early as possible (i.e. after l10n/a11y/network).

For now the whole mechanism is somewhat crude:
* download the file from the given URL
* check if it's valid yaml (we could additionally validate against [this](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-schema.html), but I'd do that in a separate PR)
* save the file to a temporary directory
* invoke `sudo` to move it to `/` on the live system (the ubuntu user doesn't have write-access, but can run `sudo` passwordless)
* restart subiquity via `POST /meta/restart`
* ~relaunch the UI via `execv(Platform.resolvedExecutable, [])`~
* now resets the UI be rebuilding the `InstallerWizard`

~P.S. Test failure that somehow made its way into main is handled in #482~

Close #483 
UDENG-2452


Feichtmeier:
((( little UI feedback: the YaruCircularProgressIndicator should prbly be colored white if put into the green button :) )))

d-loose:
> ((( little UI feedback: the YaruCircularProgressIndicator should prbly be colored white if put into the green button :) )))

UI is still very tentative, but good point - added [here](https://github.com/canonical/ubuntu-desktop-provision/compare/3a8c1b4deabbb79f452e92ac35a1eb62aa916bc9..190b9eb1d46892f0d1da8e01bb13d83830adbe35)

packages/subiquity_client/lib/src/client.dart
```diff
@@ -577,6 +577,21 @@ class SubiquityClient {
...
+
+  Future<void> restart() async {
+    final request = await _openUrl('POST', 'meta/restart');
+    try {
+      await request.close();
+    } on HttpException catch (e) {
+      if (e.message
+          .contains('Connection closed before full header was received')) {
+        _log.info(
+            'Subiquity closed the connection before the response was received. This is expected.');
```
spydon, line 589: What a strange behavior

packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -597,6 +597,10 @@ void main() {
         AdJoinResult.UNKNOWN,
       );
     });
+
+    test('restart', () async {
+      await client.restart();
```
spydon, line 602: Should this use the `returnsNormally` matcher maybe?

packages/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_model.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:safe_change_notifier/safe_change_notifier.dart';
+import 'package:stdlibc/stdlibc.dart';
+import 'package:subiquity_client/subiquity_client.dart';
+import 'package:ubuntu_bootstrap/ubuntu_bootstrap.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:yaml/yaml.dart';
+
+final _log = Logger('autoinstall_model');
+
+final autoinstallModelProvider = ChangeNotifierProvider(
```
spydon, line 18: We should try to not introduce more `ChangeNotifierProvider`s since they are discouraged by Riverpod:
https://riverpod.dev/docs/providers/change_notifier_provider

But maybe we should just convert all of them at a later stage?

packages/ubuntu_bootstrap/lib/widgets/option_button.dart
```diff
@@ -1,8 +1,7 @@
 import 'package:flutter/material.dart';
-import 'package:ubuntu_bootstrap/pages/try_or_install/try_or_install_model.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
-class OptionButton extends StatelessWidget {
```
spydon, line 5: Maybe it actually is trivial to convert those tile widgets that I created to use this one...

spydon (approved):
Some small nits, but overall looks good to me!

packages/subiquity_client/lib/src/client.dart
d-loose: Not sure if there's something weird in dart's http client, I didn't investigate it further

packages/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_model.dart
d-loose: I'm aware of that, but I'd rather stick with a reasonably consistent architecture for now. I'd really like to improve all of this at some point :)

packages/subiquity_client/lib/src/client.dart
spydon: I bet the weird behavior is on subiquity's side, that they close down everything before they got the ack

spydon (approved):

d-loose:
@spydon I went with the simple solution for the UI reset for now. Waiting for subiquity to restart still happens on the autoinstall page.
It got a bit messy when I tried to move things around and show the loading screen while subiquity is (re-)starting. Ran into some confusing race conditions caused by some widget rebuilds I couldn't wrap my head around today. Might come back to this another day, if we have time left to improve this.

spydon (approved):
Lgtm, let's try to improve on the restart functionality some other day, maybe when we go over all of the providers

d-loose:
Merging this now as is. Will probably update copy in a second pass.

PR #488, opened 2024-03-07
feat(snap): update provd socket paths

matthew-hagemann:
This aligns the snap with the socket path that will be used for the init backend moving forward
I had to move it into `/run/gnome-initial-setup` as the gis user has ownership only over that `/run` subdirectory


d-loose (approved):
LGTM :+1:

PR #490, opened 2024-03-08
feat: Passphrase page

spydon:
image

I just noticed that there are 3 different almost identical widgets for storing passwords/passphrases, I'll consolidate these in a later PR since this PR already is way too big.


melos.yaml
d-loose, line 1: If it's not too much to ask, I'd be very happy if we could separate the yaru migrations into its own PR (or at the very least its own commit) :).

spydon: I thought I was doing that, but then I realized it was in the middle of some uncommitted changes. :sweat_smile: 
Probably easier to do it again in its own PR and then rebase this one on top of that, it was not that much changes, mostly just imports.

packages/ubuntu_bootstrap/lib/installer/installer_wizard.dart
```diff
@@ -75,7 +75,9 @@ class _InstallWizard extends ConsumerWidget {
         ].contains(route)) {
           return true;
         } else {
-          return ref.read(installerModelProvider).hasRoute(route);
+          return context.mounted
+              ? ref.read(installerModelProvider).hasRoute(route)
+              : false;
```
d-loose, line 80: As mentioned yesterday, this doesn't solve the issue in `installer_test.dart` but merely moves it to a different place, see [here](https://github.com/canonical/ubuntu-desktop-provision/actions/runs/8202230324/job/22432561368?pr=490#step:7:375).
It first appeared after #451, which we need to partially [revert](https://github.com/canonical/ubuntu-desktop-provision/pull/484/commits/94c2c8b241ac9a76c16d61fc23e36f7ed88a55fe#diff-537df1c6577cdeacdbe4f4c502f6f51a573c63eb6ecb24f6df9571519b2a580f) to enable the UI reset for the autoinstall page.

packages/ubuntu_bootstrap/lib/services/storage_service.dart
```diff
@@ -44,14 +44,14 @@ class StorageService {
...
-  set securityKey(String? securityKey) {
+  /// A passphrase for full disk encryption.
+  String? get passphrase => _passphrase;
+  set passphrase(String? securityKey) {
     if (securityKey != null) {
       final hiddenKey = '*' * securityKey.length;
       _log.debug('set security key: $hiddenKey');
     }
-    _securityKey = securityKey;
+    _passphrase = securityKey;
```
d-loose, line 54: `s/securityKey/passphrase/g`

melos.yaml
spydon: Created #491 and rebased on that one :)

d-loose (approved):
LGTM :+1:

PR #493, opened 2024-03-08
feat: IdentityPage with new design

spydon:
image


d-loose (approved):
Awesome, LGTM :+1:
(I suspect the dependency issues are due to `ubuntu_utils` missing in a pubspec file, so melos doesn't find it)

spydon:
> Awesome, LGTM 👍 (I suspect the dependency issues are due to `ubuntu_utils` missing in a pubspec file, so melos doesn't find it)

They all had it in their pubspecs, and weirdly it passed after a re-run :man_shrugging:

PR #494, opened 2024-03-11
fix: Rename `PreviousWizardButton` to `BackWizardButton`

spydon:
And the label is now "Back" instead of "Previous".


d-loose (approved):
LGTM :+1:

PR #503, opened 2024-03-12
feat: New image assets and svg color mapping

spydon:
This PR adds the new image assets and sets up color mapping so that the custom accent colors are used instead of orange in the images.


packages/ubuntu_provision/lib/src/providers/page_images.dart
```diff
@@ -91,29 +115,87 @@ class PageImages {
...
+
+  Color? _accentColor({bool darkMode = false}) => darkMode
+      ? themeVariantService.themeVariant?.darkTheme?.primaryColor ??
+          themeVariantService.themeVariant?.theme?.primaryColor
+      : themeVariantService.themeVariant?.theme?.primaryColor;
+
+  bool get _hasUniqueAccentColors =>
+      themeVariantService.themeVariant?.darkTheme?.primaryColor !=
+      themeVariantService.themeVariant?.theme?.primaryColor;
+}
```
d-loose, line 175: Would it possible to take the accent color directly from the theme of the current build context instead? Currently this only works with an explicitly specified accent color in the config (i.e. it doesn't work if you just set a custom flavor in the config).

spydon: Aah, updating the flavor doesn't update the theme variant? The problem is that we don't have the build-context here, and it gets a bit dirty as soon as buildcontexts have to be passed in to services. One possible solution is to create a provider for the theme and use that similarly to the brightness provider.

spydon: Since we're moving the precaching into the loading provider (already done for stage 3) we don't have a context there to pass, and generally one should avoid passing contexts over asynchronous gaps.
I'll have a look at how terrible it would be to create a theme provider. :sweat_smile:

spydon: I tried implementing a theme provider or something similar, but it got extremely messy.
If we just listen to the flavor provider too, like in https://github.com/canonical/ubuntu-desktop-provision/pull/503/commits/cdcd9e03555fe8db6862966c642085ab9bb682a4 we would cover all use-cases right?

d-loose: Looks like a much cleaner solution!

Something's broken now when I define a flavor in the config though:
```
flutter: ERROR ubuntu_bootstrap: Unhandled exception
	_AssertionError: 'package:flutter/src/widgets/framework.dart': Failed assertion: line 5205 pos 12: '!_dirty': is not true.
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      Element.rebuild (package:flutter/src/widgets/framework.dart:5205:12)
#3      ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:5462:5)
#4      ComponentElement.mount (package:flutter/src/widgets/framework.dart:5456:5)
#5      _UncontrolledProviderScopeElement.mount (package:flutter_riverpod/src/framework.dart:315:11)
#6      Element.inflateWidget (package:flutter/src/widgets/framework.dart:4335:16)
...
```

Will have a closer look tomorrow morning

spydon: Oh boy, that was one wild ride! At some point we should really clean up the root structure of the apps. :sweat_smile: 

Now it's solved in https://github.com/canonical/ubuntu-desktop-provision/pull/503/commits/5b06ad6344b9e47e548d68b7f1ef41830b1dbbbc by using a `ProviderContainer` instead, I think it got a bit cleaner than previously at least.

spydon: Haha, that was not the end of the journey! Apparently adding a `ProviderContainer` as the `parent` field of the `ProviderScope` was deprecated just a few weeks ago. So now I went with creating a service for it instead (https://github.com/canonical/ubuntu-desktop-provision/pull/503/commits/9e12a7b32f4bc542e3de10918ddc75da79222053), since it really needs to be there before the app really starts. At some point in the future maybe we can convert it to a `FutureProvider`, but then we need to do some other rearrangements.

d-loose: Hm, could we at least keep the provider (as a simple wrapper around the service), so that the pages don't have to directly call `getService()` in their build functions?

d-loose (approved):
Really nice! LGTM :+1:

PR #504, opened 2024-03-13
feat: Configurable visibility in whitelabel.yml

spydon:


packages/ubuntu_bootstrap/lib/installer.dart
```diff
@@ -108,6 +109,9 @@ Future<void> runInstallerApp(
     () => PageConfigService(
       config: tryGetService<ConfigService>(),
       includeTryOrInstall: includeTryOrInstall,
+      allowedToHide: InstallationStep.values
+          .where((e) => e.allowedToHide)
+          .map((e) => e.route),
```
d-loose, line 114:
```suggestion
          .map((e) => e.name),
```

`_excludedPages` in `PageConfigService` uses the names of the pages without the leading '/'

README.md
```diff
@@ -82,9 +82,12 @@ theme:
 # - privacy: Enable location services
 # - timezone: Set the timezone
 # - telemetry: Enable sending telemetry
+#
+# Do note that only some non-mandatory pages offer the possibility to be hidden.
```
d-loose, line 86: Should we name the pages that can be hidden explicitly?

spydon: Right, I forgot to circle back on that, which ones do we want to be able to hide?
Are accessibility, source, try-or-install and refresh enough for stage 3?

packages/ubuntu_bootstrap/lib/installer.dart
spydon: Wtf... How did this even work...? :sweat_smile: 
Good catch!

d-loose (approved):

PR #509, opened 2024-03-14
chore: Update to Melos v5.1.0

spydon:


d-loose (approved):
Doesn't seem to fix the random dependency issue :(
But probably good to upgrade anyways!

PR #513, opened 2024-03-14
fix: Only display the TPM option if TPM is available

### LINKED ISSUE #150: Experimental options such as TPM need to visibly reflect subiquity return values
Raised from https://bugs.launchpad.net/subiquity/+bug/2037234

```
2023-09-24 22:23:24.269117 DEBUG subiquity_client: ==> getGuidedStorageV2(true) {"status": "DONE", "error_report": null, "configured": null, "targets": [{"disk_id": "disk-sda", "allowed": ["DIRECT", "LVM", "LVM_LUKS", "ZFS"], "disallowed": [], "$type": "GuidedStorageTargetReformat"}, {"allowed": ["MANUAL"], "disallowed": [], "$type": "GuidedStorageTargetManual"}]}

The client should not have shown you the TPM encryption option in this case (it doesn't look as if the ISO even suports this as an option).
```

On ubuntu-budgie-installer the TPM experimental option is visible and able to be selected; subiquity says "disallowed" - so the option should be invisible/disabled ?

spydon:
Closes #150

### DIFF
--- a/packages/ubuntu_bootstrap/lib/pages/storage/storage_dialogs.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/storage/storage_dialogs.dart
@@ -88,10 +88,11 @@ Future<void> showAdvancedFeaturesDialog(
                       onChanged: (v) => guidedCapability.value = v!,
                     ),
                   ],
-                  TpmOption(
-                    guidedCapability: guidedCapability,
-                    model: model,
-                  ),
+                  if (model.hasTpm)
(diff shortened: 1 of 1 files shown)


spydon:
Just noticed that it should already be disabled (instead of removed), so this can be closed.
https://github.com/canonical/ubuntu-desktop-provision/blob/main/packages/ubuntu_bootstrap/lib/pages/storage/storage_dialogs.dart#L151-L153

PR #522, opened 2024-03-15
docs: Correct format for accent color in whitelabel.yaml

eeickmeyer:
Found out the hard way that whitelabel.yaml will treat the # as a comment, but will honor the accent color as chosen by the hex code without the #.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,11 @@ app-name: <string>
 # (Optional) Overrides the theme's accent colors (remember the quotes)
 theme:
   light:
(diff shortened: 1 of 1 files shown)


spydon (approved):

eeickmeyer:
Honestly, I take it back, I just re-did my whitelist.yaml with `"#______"` where the underscores are the hex code and it worked, so I guess we just need to make that more clear. I skimmed over that part.

spydon:
Oooh, I thought you changed the docs the other way around, sloppy sunday reviewing on my part!.

PR #523, opened 2024-03-18
fix: PageImages should be able to load SVGs from file

spydon:
It was reported from the flavor folks that SVGs couldn't be loaded from the filesystem anymore, this fixes that and makes the test a bit more reliable.


d-loose (approved):
LGTM :+1:

Issue #495, opened 2024-03-11
Collection of UI issues


d-loose:
- [x] Remove back button on first page
- [x] Change title of accessibility page
- [x] Update copy on keyboard page
- [x] Review margins on network page
- [x] Update mascot on refresh page
- [x] Update copy on refresh page
- [x] Replace logo with mascot on try-or-install page
- [x] Update copy on try-or-install page
- [x] Select 'install' by default on try-or-install page
- [x] Change button label to 'close' when selecting 'try' on try-or-install page
- [x] Check alignment on try-or-install page
- [x] Update copy on autoinstall page
- [x] Check alignment and font size on autoinstall page
- [x] Update copy on source selection page
- [x] Check alignment on drivers/codecs page
- [x] Update copy on drivers/codecs page
- [x] Remove boxes on storage page, separate first two options
- [x] Remove release number on storage page (?)
- [x] Add 'experimental' badge to TPM option in advanced feature dialog
- [x] Implement new designs for guided resize and guided reformat pages
- [x] Implement new AD page desig
- [x] Implement error page design
- [x] Update copy on RST and bitlocker pages
- [x] Change to elevated button on RST and bitlocker pages
- [x] Interchange buttons on 'done' page
- [x] Update copy on 'done' page

anasereijo:
remove back button

<img width="932" alt="Screenshot 2024-03-13 at 15 09 29" src="https://github.com/canonical/ubuntu-desktop-provision/assets/36884067/fca2e036-a2ed-40c6-975d-33cf5813e668">

anasereijo:
<img width="951" alt="Screenshot 2024-03-13 at 15 12 08" src="https://github.com/canonical/ubuntu-desktop-provision/assets/36884067/b3552793-a141-4d0b-9850-41452713efa2">
change title to 'Accessibility'

anasereijo:
<img width="601" alt="Screenshot 2024-03-13 at 15 14 30" src="https://github.com/canonical/ubuntu-desktop-provision/assets/36884067/9812af6e-b044-4f4a-b471-60c1a47858a0">

change to:
<img width="422" alt="Screenshot 2024-03-13 at 15 14 49" src="https://github.com/canonical/ubuntu-desktop-provision/assets/36884067/220f296f-98eb-4fd6-9e24-bea87c76c02b">

anasereijo:
[...]

PR #532, opened 2024-03-18
fix(bootstrap): remove horizontal spacer on storage page

d-loose:
Sorry, missed that one. With the boxes removed we don't need any additional spacers.

Ref: #495
UDENG-2506


spydon (approved):
Ah, missed that one too

PR #535, opened 2024-03-18
chore: Clean up translations and add melos action to do it

spydon:


d-loose (approved):
Thanks, that's very useful!

PR #546, opened 2024-03-20
feat: update mascot on try-or-install and slides

d-loose:
Ref: #495
UDENG-2506

Screenshot from 2024-03-20 19-04-19
Screenshot from 2024-03-20 19-04-06


spydon (approved):
Nice! I thought it would look bad, but this is great!

PR #548, opened 2024-03-21
feat(init): add provd GDM service

d-loose:
Dart client for provd-based GDM service from #455

Edit: Go Code sanity failure seems to be because I've cherry-picked only the proto files here, but not the generated go files. Should be fine once rebased onto #455.


matthew-hagemann:
@d-loose I see there is a line in the snapcraft.yaml we will need to change which references the old gdm filename:

https://github.com/canonical/ubuntu-desktop-provision/blob/ci/init/snap/snapcraft.yaml#L76

d-loose:
Oh, good catch! We won't need this patch anymore when provd handles this

PR #551, opened 2024-03-21
feat: New `ErrorPage` design

spydon:
This also prepares for ubuntu_init to use it, but doesn't convert it over since there are a few more things to sort out for that.

Screenshot from 2024-03-21 19-20-05


d-loose (approved):
Very nice!

There's some mild scrollbar weirdness going on:

[Screencast from 2024-03-21 19-51-26.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/113362648/43b3ed17-233b-432e-a826-0cc22f2f564b)

But I think it's fair to treat this as a very minor bug and fix it separately :)

LGTM :+1:

PR #552, opened 2024-03-21
fix: More title space on the done page

spydon:
Some long distribution titles overflowed apparently,

image


d-loose (approved):
Thanks, good catch!

PR #557, opened 2024-03-22
Revert: feat(bootstrap): update application icon

eeickmeyer:
Reverts 866e489460a582ecdfc47df22f4e7e581e2a0913

This commit overrides any flavor-specific changes and `dpkg-dirvert`s done to the `ubiquity` icon within various icon themes, by various flavors to rebrand the "Install RELEASE" where RELEASE is the flavor name, in particular the yaru theme. If you wish to update the icon, please make a pull request to https://github.com/ubuntu/yaru.

Therefore, what the commit did is overrode flavor branding, which is a regression.


eeickmeyer:
Basically, this is my way of saying the commit in quesiton is a hard NACK from myself as a flavor lead. See screenshot below:

VirtualBox_ISOTest_22_03_2024_16_02_42

This cannot be simply switched-out as it would require significant changes to `casper` since the icon is part of the snap, and editing the `.desktop` file is not possible as that is immutable. The flavors have, historically, relied on the `ubiquity` icon in the past being the de-facto installer icon which is replacable via `dpkg-divert` and expected in icon themes, such as Yaru.

eeickmeyer:
For reference, here's what we expect for Ubuntu Studio and Edubuntu respectively, and that due to the nature of snaps we cannot override the icon that was used:

ubiquity-papirusubiquity

spydon (approved):
Lgtm, the icon will be added through distro patching from our side instead

spydon:
@eeickmeyer your commit isn't signed so this can't be merged, do you want to fix that and we'll merge this or should I open up another one from my end?

spydon:
Nvm, I could sign it for you

PR #558, opened 2024-03-25
fix: realName on identity page can’t be empty.

### LINKED ISSUE #540: Empty real name on init frontend
We received an issue from QA that, according to journal log, when realname (Your name) is empty, user creation can be stuck.

Related journal entries:
Mar 18 18:19:20 localhost gnome-initial-setup.desktop[4207]: flutter: ERROR ubuntu_init: Unhandled exception
Mar 18 18:19:20 localhost gnome-initial-setup.desktop[4207]: GrpcError: gRPC Error (code: 3, codeName: INVALID_ARGUMENT, message: received an empty realName, details: [], rawResponse: null, trailers: {})

Related line:
https://github.com/canonical/ubuntu-desktop-provision/blob/7bf81d2af29cea5737e747a204859487cc8a1a52/provd/internal/services/user/user.go#L143

BLKKKBVSIK:
In this PR:

- fix #540

### DIFF
--- a/packages/ubuntu_provision/lib/src/identity/identity_model.dart
+++ b/packages/ubuntu_provision/lib/src/identity/identity_model.dart
@@ -134,7 +134,8 @@ class IdentityModel extends SafeChangeNotifier with PropertyStreamNotifier {
 
   /// Whether the current input is valid.
   bool get isValid {
-    return realName.length <= kMaxRealNameLength &&
+    return realName.isNotEmpty &&
+        realName.length <= kMaxRealNameLength &&
         hostname.isNotEmpty &&
         hostname.length <= kMaxHostnameLength &&
         username.isNotEmpty &&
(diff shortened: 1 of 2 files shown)


packages/ubuntu_provision/lib/src/identity/identity_model.dart
```diff
@@ -134,7 +134,8 @@ class IdentityModel extends SafeChangeNotifier with PropertyStreamNotifier {
 
   /// Whether the current input is valid.
   bool get isValid {
-    return realName.length <= kMaxRealNameLength &&
+    return realName.isNotEmpty &&
```
spydon, line 137: Does subiquity not allow an empty real name? Because `/etc/passwd` does allow empty real names.

spydon: Ah, I just read the issue, so it gets some problem on the way, I guess it's fine then!
The test checking that the real name should be allowed to be empty has to be updated though.

spydon (approved):

packages/ubuntu_provision/lib/src/identity/identity_model.dart
medicalwei: In Subiquity it seems like it is handling empty realname by copying username to realname when it's empty:

https://github.com/canonical/subiquity/blob/main/subiquity/models/identity.py

PR #568, opened 2024-03-27
fix: The product name should come from SMBIOS 'Product Name'

vanvugt:
Unless you are LENOVO or IBM in which case they have forever put their product names in the 'Version' field instead.

This fixes the default hostname being nonsensical on non-Lenovo systems.

Closes: https://bugs.launchpad.net/bugs/2056167
Closes: https://github.com/canonical/ubuntu-desktop-installer/issues/2394


spydon (approved):
Super, thanks!

PR #572, opened 2024-03-27
feat: Ubuntu pro page

BLKKKBVSIK:
In this PR:

- All the Ubuntu Pro pages.


spydon:
It looks like you have a faulty rebase or something similar going on here.

BLKKKBVSIK:
Yeah ! Trying to fix this. It got a bit messy 😅

spydon:
> Yeah ! Trying to fix this. It got a bit messy 😅

Ping me if you need help fixing it tomorrow 😄

BLKKKBVSIK:
> > Yeah ! Trying to fix this. It got a bit messy 😅
> 
> Ping me if you need help fixing it tomorrow 😄

I think I'll need some help.
My tests aren't passing since the merge, lots have change in ubuntu_init

d-loose:
The integration tests are failing because there's no fake service for ubuntu pro to drive those tests (`fake_init.dart` in `ubuntu_provision_test`).

The e2e test tries to go through all of the ubuntu pro pages, which cannot work unless you actually try to pro-attach the runner (which probably isn't a great idea). So the test should just select 'skip' on the initial pro page.

Generally, the commit history is still quite messy and contains a lot of unnecessary things. Would it be possible to remove all of those duplicate commits and merges from the main branch? Ideally it should only include my initial implementation of the service/model and your UI, and be rebased onto main.

packages/ubuntu_init/lib/src/ubuntu_pro/ubuntu_pro_onboarding_page.dart
```diff
@@ -0,0 +1,75 @@
...
+      windowTitle: l10n.ubuntuProPageTitle,
+      title: '',
+      bottomBar: WizardBar(
+        leading: const BackWizardButton(),
+        trailing: [
+          WizardButton(
+            label: UbuntuLocalizations.of(context).nextLabel,
+            onActivated:
+                model.selection == UbuntuProOnboardingPageSelection.skipForNow
+                    ? () => Wizard.of(context).jump(InitStep.privacy.route)
```
d-loose, line 33: Could you leave a todo note here? I'd like to get rid of this explicit jump, since it makes assumptions about the following pages. Instead we should handle this in the `load` methods of the `UbuntuProPage` and `UbuntuProSuccessPage`, by checking for some sort of 'skip' property.

```suggestion
                    ? () => Wizard.of(context).jump(InitStep.privacy.route)
                    // TODO: Handle page skipping in load methods
```

d-loose: This is actually why the `init_wizard_test` currently gets stuck [here](https://github.com/canonical/ubuntu-desktop-provision/blob/9d659cf1fd0be48968ed374bc07c9443255c933f/packages/ubuntu_init/test/init_wizard_test.dart#L226). It leaves the option button on the `UbuntuProOnboardingPage` unchanged on "Skip", so that the wizard tries to advance to the `PrivacyPage` when pressing "Next". But the privacy page isn't included in that test, so it gets stuck.

PR #574, opened 2024-03-28
Chown settings for copy worker

matthew-hagemann:
Workaround to a permission bug in GIS that blocks the copy worker from running:
https://bugs.launchpad.net/ubuntu/+source/gnome-initial-setup/+bug/2059121

I broke this out into its own service to make it easier to remove down the line when the GIS bug is fixed


provd/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/canonical/ubuntu-desktop-provision/provd
 
-go 1.22.0
+go 1.21.7
```
spydon, line 3: It downgrades go?

spydon (approved):
Lgtm

provd/go.mod
matthew-hagemann: Yea, I ran into issues there on the runners, let me give it another look and see if I can fix it :slightly_smiling_face:

PR #578, opened 2024-04-02
feat: New design for the factory reset tools

spydon:
[Screencast from 2024-04-02 18-51-07.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/744771/e7f1914b-375a-47ab-af48-de9e211bc3c1)

Yaru will be swapped back to a published version in a few hours once the new yaru version is published.


packages/factory_reset_tools/lib/dbus/drive_manager.dart
```diff
@@ -0,0 +1,203 @@
+import 'dart:async';
+
+import 'package:dbus/dbus.dart';
+import 'package:factory_reset_tools/dbus/drive_data.dart';
+import 'package:factory_reset_tools/dbus/reset_media.dart';
+
+const int minimumRequiredDiskSize = 12 << 30;
```
d-loose, line 7: This is 12GiB, but in the loop further down the comment says:
```
// 2. has enough Size (we assume 16G)
```
Is one of those a typo?

d-loose (commented):
We could avoid a little bit of duplication by depending on `ubuntu_provision` - is there a reason for not doing that?

packages/factory_reset_tools/lib/dbus/drive_manager.dart
```diff
@@ -0,0 +1,203 @@
...
+          driveObjectPath,
+          blockDeviceObjectPath,
+        ),
+      );
+    }
+    drives.sort((a, b) => a.id.compareTo(b.id));
+    return drives;
+  }
+
+  static Stream<ResetMediaCreationProgress> dummyCreateResetMedia(
```
d-loose, line 176: Nitpick: `s/dummy/fake`

packages/factory_reset_tools/lib/dbus/drive_manager.dart
spydon: Probably a typo, right @medicalwei?

spydon:
> We could avoid a little bit of duplication by depending on `ubuntu_provision` - is there a reason for not doing that?

Yeah, the duplication is the `HorizontalPage` (the rest I moved to our terribly unorganized ubuntu_utils package :sweat_smile:), but we don't want this interconnected with the pageImagesProvider here.
We could refactor them, but I think that could probably be future task?

packages/factory_reset_tools/lib/dbus/drive_manager.dart
medicalwei: It was a consideration that disks might be actually smaller than 16GiB or spaces taken by partition metadata. (eg. being 16,000,000,000 bytes instead of 17,179,869,184 bytes.)

But it's fine that you can reimplement that part.

d-loose:
Ah right, the page images make it a bit tricky. Future task it is :)

packages/factory_reset_tools/lib/dbus/drive_manager.dart
medicalwei: Thanks, it was my placeholder code for testing the UI

packages/factory_reset_tools/lib/dbus/drive_manager.dart
medicalwei: It is also worthy to mention that, the mockup does not hide the disks that does not meet the size requirement, they are just greyed out.

spydon: Is 12 enough? Then I think we should keep that

medicalwei: For now yes, as the disk partition we reserve for factory reset is 12GiB

fourdollars:
Ideally the minimum size shall be decided by the real storage usage in the reset partition.
For example, the storage usage in the reset partition used 3.2GB so the minimum required size shall be 3.2*1.1 = 3.52 GB.
That means a 4GB USB pendrive is OK for this case.

packages/factory_reset_tools/lib/dbus/drive_manager.dart
spydon: So even 4GB should be enough according to @fourdollars, do you think it would be fine if I lowered it that low @medicalwei?

d-loose (approved):
LGTM :+1:

packages/factory_reset_tools/lib/dbus/drive_manager.dart
medicalwei: IMO @fourdollars says the minimum size should be according to the actual used size of the reset partition, and the number he proposed is just for example.

spydon: Aah I see, do we know the approximate size of the reset partition?

PR #581, opened 2024-04-03
ci: Add snap directory for the reset tools

spydon:
No idea if these actually work yet...


ci/snap/reset-tools/snapcraft.yaml
```diff
@@ -0,0 +1,138 @@
...
+      dart pub global run melos bootstrap
+      cd packages/factory_reset_tools
+      flutter build linux --release -v
+      cp -r build/linux/*/release/bundle/* $CRAFT_PART_INSTALL/bin/
+
+  factory-reset-tools-cli:
+    after: [flutter-git]
+    source: .
+    plugin: nil
+    flutter-target: lib/main.dart
```
d-loose, line 104:
```
Bad snapcraft.yaml content:
- extra field 'flutter-target' not permitted in 'parts.factory-reset-tools-cli' configuration 
```

spydon: I just got it from here: https://github.com/canonical/factory-reset-tools/blob/main/snap/snapcraft.yaml
But it looks to me like that line might not be needed at all?

ci/snap/reset-tools/local/factory-reset-tools.desktop
```diff
@@ -0,0 +1,5 @@
+[Desktop Entry]
+Exec=factory-reset-tools
+Name=Factory Reset Tools
+Type=Application
+Icon=${SNAP}/meta/gui/factory-reset-tools.png
```
spydon, line 5:
```suggestion
Icon=${SNAP}/meta/local/factory-reset-tools.png
```

ci/snap/reset-tools/snapcraft.yaml
spydon, line 104:
```suggestion
```

d-loose: ah, that was using snapd's flutter plugin (which I've never used, so I don't know how well this works)

spydon: @d-loose Should we just merge this and iterate on it later, once the tools are ready?
So that the job doesn't fail.

d-loose (approved):

PR #582, opened 2024-04-03
feat(init): migrate to provd based telemetry service

d-loose:
Removes the sysmetrics package and corresponding telemetry service in favor of a provd based service.

~~e2e test and go code check are expected to fail as the provd implementation is still missing.~~

Ref #363
UDENG-2204


d-loose:
Errors returned by provd are ignored for now. I'd merge the UI update from #667 first and think about error handling afterwards.

ci/snap/init/snapcraft.yaml
d-loose, line 1: @matthew-hagemann I suppose we can remove this library from the snap since we're now using the native Go API

spydon (approved):
Nice, lgtm!

ci/snap/init/snapcraft.yaml
d-loose: .. and since the entire code for this moved to the deb package anyways.. :D

PR #583, opened 2024-04-03
ci: only build debs on push's to main

matthew-hagemann:


.github/workflows/ci.yml
```diff
@@ -255,6 +255,7 @@ jobs:
           token: ${{secrets.CODECOV_TOKEN}}
 
   build-and-upload:
+    if: github.event_name == 'push'
```
spydon, line 258: I don't think that this will help? Since it still active on all PR branches that targets main?

spydon: Or will the event_name be `pull_request` for all pushes to a PR branch?

matthew-hagemann: You mean all current branches that are open? We could rebase / cherry pick until they are all closed out?

spydon: I meant for future branches too, the current branches I wouldn't worry about too much.
Let's merge this and see what happens. :smile:

spydon (approved):

.github/workflows/ci.yml
matthew-hagemann: > Or will the event_name be `pull_request` for all pushes to a PR branch?

Ah yes, so from what I understand the CI triggers on PR's and Pushes to main. I believe a push to a feature branch that has an open PR triggers the the PR flow, not the Push flow.

Lets see, easy enough to revert :smile:

PR #584, opened 2024-04-04
feat: Add Done page to reset media creation

spydon:
image

Also (in separate commits):
 * Add i10n for the progress state
 * Error handling when `ResetMediaCreationStatus` ends up in `failed`


packages/factory_reset_tools/lib/dbus/reset_media.dart
```diff
@@ -10,7 +11,22 @@ enum ResetMediaCreationStatus {
   copying,
   finalizing,
   finished,
-  failed
+  failed;
+
+  String displayName(FactoryResetToolsLocalizations lang) {
```
d-loose, line 16: nitpick: would be slightly more readable with a switch expression

packages/factory_reset_tools/lib/pages/creation_progress_page.dart
```diff
@@ -24,6 +25,11 @@ class _CreationProgressPageState extends ConsumerState<CreationProgressPage> {
   );
 
   void onStatusChanged(ResetMediaCreationProgress progress) {
+    if (progress.status == ResetMediaCreationStatus.finished) {
+      Future.delayed(const Duration(seconds: 2), () {
```
d-loose, line 29: Is that delay intentional or left over from debugging?

packages/factory_reset_tools/lib/pages/done_page.dart
```diff
@@ -0,0 +1,60 @@
...
+                  style: theme.textTheme.headlineSmall,
+                  textAlign: TextAlign.center,
+                ),
+                Text(
+                  lang.resetMediaReadyBody,
+                  textAlign: TextAlign.center,
+                ),
+                SizedBox(
+                  width: 100,
+                  child: FilledButton(
```
d-loose, line 45: I'm no designer, but I think an `ElevatedButton` would be more appropriate here

packages/factory_reset_tools/lib/pages/creation_progress_page.dart
spydon: Intentional, otherwise you never see the last state 😅

packages/factory_reset_tools/lib/dbus/reset_media.dart
spydon: True! I always forget about new features when I'm coding on autopilot 😄

packages/factory_reset_tools/lib/pages/done_page.dart
spydon: It does look a lot better, even though it is not according to the design, but on the other hand I barely followed the design at all because it wasn't ready. Hopefully the designer that reviews this won't kill me. :sweat_smile:

d-loose (approved):
LGTM :+1:

PR #585, opened 2024-04-04
fix(bootstrap): reset storage model when going back

d-loose:
This ensures that the storage model is reset if the user decides to go back in the flow and select a different option.

The necessary callback was already present but somehow ended up in the *next* button (probably when we switched from the abstract "wizard actions" to buttons in https://github.com/canonical/ubuntu-desktop-installer/pull/1954).
I've moved them into the back buttons and added corresponding tests.


d-loose:
The underlying issue is a different one - will open a separate PR with a proper fix

PR #587, opened 2024-04-04
fix(bootstrap): reset storage model when navigating back

d-loose:
This ensures that the storage model is reset if the user decides to go back in the flow and select a different option.

I've tried to clarify the meaning of `NextWizardButton`'s `onReturn` parameter (which was previously called `onExecute` and `onBack` before that). We should definitely spend some time on refactoring `ubuntu_wizard` in the future, since some of the mechanisms involved are quite tricky and hard to understand.

Another issue is that there are multiple ways of modifying the behavior of the buttons in the bottom bar of the new `HorizontalPage`: either by overriding the entire `bottomBar` or by passing arguments such as `isNextEnabled` that modify the *default* bar. I've added an assertion to assure that those parameters aren't specified at the same time. I'd rather remove one of those options in the future, though.

### DIFF
--- a/packages/ubuntu_wizard/lib/src/wizard_button.dart
+++ b/packages/ubuntu_wizard/lib/src/wizard_button.dart
@@ -218,7 +218,7 @@ class NextWizardButton extends StatelessWidget {
     this.highlighted = false,
     this.arguments,
     this.onNext,
-    this.onExecute,
+    this.onReturn,
   });
 
   final String? label;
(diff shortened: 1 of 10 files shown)


packages/ubuntu_wizard/lib/src/wizard_button.dart
```diff
@@ -228,7 +228,10 @@ class NextWizardButton extends StatelessWidget {
   final bool highlighted;
   final Object? arguments;
   final WizardCallback? onNext;
-  final WizardCallback? onExecute;
+
+  /// Called when returning to the current route after navigating to the next one.
+  /// (e.g. when pressing the back button on the next page)
+  final WizardCallback? onReturn;
```
spydon, line 234: It's weird that the button is responsible for this... But that is for a later refactor.

spydon (approved):
Lgtm

packages/ubuntu_wizard/lib/src/wizard_button.dart
d-loose: Yes, it's quite ugly too carry around a chain of coroutines like this - and terribly annoying to debug

PR #589, opened 2024-04-05
feat(bootstrap): show loading page while waiting for subiquity

d-loose:
This moves the futures that wait for the connection to subiquity and initialize the services into a future builder wrapping the `InstallerWizard`, so that we can show the loading page early on, avoiding common 'blank window' issues like https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2056160/.


packages/ubuntu_bootstrap/lib/installer.dart
```diff
@@ -172,16 +173,16 @@ Future<void> runInstallerApp(
...
-
     runApp(ProviderScope(
       child: _InstallerApp(
         theme: theme,
         darkTheme: darkTheme,
         themeVariant: themeVariant,
         windowTitle: windowTitle,
         flavor: flavorService.flavor,
+        init: getService<SubiquityServer>()
+            .start(args: subiquityArgs)
```
d-loose, line 184: I've moved this call here, so that possible exceptions thrown by it are also caught in the future builder

PR #592, opened 2024-04-05
feat(bootstrap): don't check for updates in `_initInstallerApp`

d-loose:
This is a follow-up to #589, also concerning the delay during subiquity's startup.
As Olivier pointed out [here](https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2056160/comments/9), we're currently checking for snap updates before launching the installer wizard. If there's no network connection at this point, the blocking call takes 30 seconds to time out.
I think the simplest solution is to remove that call from there entirely, as it is done [later on](https://github.com/canonical/ubuntu-desktop-provision/blob/c7967e539a90ddad181dbdb1456fd51cb092fa44/packages/ubuntu_bootstrap/lib/pages/refresh/refresh_model.dart#L24) anyways.

~~Even though the refresh page is shown after the network page, the user could still choose not to connect to the internet, so I added a connectivity check to the refresh page's `load()` function - skipping it in case there's no connection.~~

### DIFF
--- a/packages/ubuntu_bootstrap/lib/installer.dart
+++ b/packages/ubuntu_bootstrap/lib/installer.dart
@@ -254,7 +254,6 @@ Future<void> _initInstallerApp(Endpoint endpoint) async {
   final services = [
     getService<InstallerService>().init(),
     tryGetService<DesktopService>()?.inhibit() ?? Future.value(),
-    getService<RefreshService>().check(),
     getService<PageConfigService>().load(),
     geo.init(),
     telemetry.init({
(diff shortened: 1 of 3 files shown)


d-loose:
> Even though the refresh page is shown after the network page, the user could still choose not to connect to the internet, so I added a connectivity check to the refresh page's `load()` function - skipping it in case there's no connection.

I'm realizing this is a bit tricky, since it makes the integration tests depend on the network connectivity of the host. I think it's ok to remove that bit again. The user can't get stuck on the refresh page, as they can just press 'skip' if the check is taking too long (due to any kind of network troubles). Also, the following page is the autoinstall page, which relies on a network connection as well. I'll open a separate issue for those things.

spydon (approved):

PR #593, opened 2024-04-05
chore: Add l10n files for the factory reset tool

spydon:


packages/factory_reset_tools/lib/l10n/factory_reset_am.arb
```diff
@@ -0,0 +1 @@
+{}
```
d-loose, line 1: Is the file name sufficient for weblate? For the firmware updater I added the `@@locale` field [explicitly](https://github.com/d-loose/firmware-updater/blob/1e928f13c38fb6a977e24ad050f2ff6873d2f842/packages/firmware_updater/lib/src/l10n/app_am.arb#L2)

spydon: Ah, maybe it's not?
Do you mayhaps have a little script that did that?

d-loose: Not a script, just a bash command from hell
```sh
for lang in $(ls ~/projects/canonical/git/ubuntu-desktop-provision/packages/ubuntu_bootstrap/lib/l10n/*.arb | sed 's/.*ubuntu_bootstrap_\(.*\)\.arb/\1/g'); do echo -e "{\n  \"@@locale\": \"$lang\"\n}" > app_$lang.arb; done
```

spydon: Fixed!

d-loose (approved):

PR #598, opened 2024-04-08
C include path fix

### LINKED ISSUE #599: All non amd64 achs fail to build on lp

matthew-hagemann:
All the non-amd64 builds are failing on LP from the following error:
```
/usr/include/glib-2.0/glib/gtypes.h:34:10: fatal error: glibconfig.h: No such file or directory
   34 | #include <glibconfig.h>
      |          ^~~~~~~~~~~~~~
```
I've added a pkg-config line to fix this in all archs

packaging building on all on lp:
https://launchpad.net/~matt-hagemann/+archive/ubuntu/provd/+packages

closes #599

### DIFF
--- a/provd/internal/services/user/password_hash.go
+++ b/provd/internal/services/user/password_hash.go
@@ -2,7 +2,7 @@
 package user
 
 /*
-#cgo CFLAGS: -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include
+#cgo pkg-config: glib-2.0
 #cgo LDFLAGS: -lcrypt -lglib-2.0
 #include "password_hash.h"
 */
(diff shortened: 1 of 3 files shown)


matthew-hagemann:
Wrote package-include instead of pkg-config in the commit message so forced pushed with that amended :sweat_smile:

provd/internal/services/user/password_hash.go
```diff
@@ -2,7 +2,7 @@
 package user
 
 /*
-#cgo CFLAGS: -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include
+#cgo pkg-config: glib-2.0 gio-2.0 gobject-2.0
```
didrocks, line 5: do you need to include the 3 of them? Also, the "ci.yml" still include a hardcoded lib path then?

matthew-hagemann:
@didrocks updated, all building in the ppa

didrocks (approved):
That’s the right fix! :)

PR #612, opened 2024-04-09
Don't modify the gnome-initial-setup user and groups

seb128:
Provd doesn't own those and shouldn't be doing those changes

I'm unsure why you are doing that, if gnome-initial-setup has the permissions needed then provd should also be able to have them without changing the groups membership? In any case it's wrong for provd to go modify an user owned by another package. If we really need to add the gnome-initial-setup user to the gnome-initial-setup group then it should be done in the gnome-initial-setup package directly


matthew-hagemann (approved):
LGTM!

d-loose:
Dealing with the CLA check separately

PR #616, opened 2024-04-10
fix: set GIO_MODULE_DIR to the local directory also for subiquity

seb128:
it is needed because the service calls to gsettings to set the keyboard layout and since the snap is classic without the environment set it will try to load the system's libdconfsettings.so which errors out because the noble version depends on a newer libglib that snap one

https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2060387

### DIFF
--- a/ci/snap/bootstrap/snapcraft.yaml
+++ b/ci/snap/bootstrap/snapcraft.yaml
@@ -17,6 +17,7 @@ apps:
     daemon: simple
     restart-condition: always
     environment:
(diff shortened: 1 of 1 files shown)


d-loose:
Superseded by #627

Since subiquity [explicitly passes](https://github.com/canonical/subiquity/blob/6db2ff8113560fa72aa4928222fb62eb23e8de19/subiquity/server/controllers/keyboard.py#L285-L291) a custom environment to the `gsettings` command, setting `GIO_MODULE_DIR` at this level doesn't work. Instead we decided to move this into the front-end.

PR #617, opened 2024-04-10
chore: Create localized slide files

spydon:
[Screencast from 2024-04-10 16-02-31.webm](https://github.com/canonical/ubuntu-desktop-provision/assets/744771/935c85f5-408c-4269-8996-120cd2798b5a)

This also creates a small utility app that creates the slide files, it could unfortunately not be a script due to that fact that dart:ui isn't available from scripts so I couldn't get the `Locale` files in the script.


d-loose (approved):
Thanks for turning this into an HTML repo :+1:

matthew-hagemann (approved):
RIP hopes and dreams of being top contributor :broken_heart:

PR #620, opened 2024-04-11
provd: resolve build issues when using sbuild

medicalwei:
Tried using sbuild to build provd, found it is using home directory as cache which sbuild does not allow.

I followed patch similar to https://bugs.launchpad.net/ubuntu/+source/calamares-settings-ubuntu/+bug/2060151 so that GOCACHE and GOMODCACHE does not use home directory.

Also, ca-certificates is not part of sbuild chroot, so it is added to Build-Depends as well.


matthew-hagemann (approved):
LGTM :+1:

PR #621, opened 2024-04-11
provd: avoid chown to gnome-initial-setup group

medicalwei:
Although gnome-initial-setup user is created in postinst, gnome-initial-setup group only exists when the installed environment has been run once (I did not find out who created it, but it might be gdm3, not sure.)

This patchset avoids chown to gnome-initial-setup group so the installation can be done in livecd-rootfs.


medicalwei:
Closed because we are fixing gnome-initial-setup to create group over there

PR #626, opened 2024-04-11
fix(provd): use provd group, not gnome-initial-setup group

matthew-hagemann:
Move away from using gnome-initial-setup group, which is owned and created by gnome-initial-setup, and make use of a provd group, including all its lifecycle management.


seb128 (approved):
Thanks!

Issue #629, opened 2024-04-11
disable validate button on auto install page when the url is wrong


anasereijo:
<img width="1066" alt="Screenshot 2024-04-11 at 14 09 31" src="https://github.com/canonical/ubuntu-desktop-provision/assets/36884067/d6b32ed3-61db-4ae4-b877-700f760401f6">

PR #631, opened 2024-04-11
fix: Only show validate button if url is valid

spydon:
Closes: #629


d-loose:
Sorry, that came out of my meeting with Ana - to provide bit more context, we should disable the button while an error message is being shown (i.e. while the form is invalid). We don't need to explicitly check for the validity of the URL.

spydon:
@d-loose alright, fixed!

d-loose (approved):

PR #632, opened 2024-04-11
feat(bootstrap): set keyboard via setxkbmap

d-loose:
Follow-up to #627 
Fixes https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2058503 by additionally invoking `setxkbmap`. It introduces a tiny amount of lag after selecting the keyboard layout, as the UI waits for the command to return. We could instead mark this as `unawaited`, since it's unlikely that the user manages to type in the test field before the command returns - any thoughts?

I've checked that this works on the latest Xubuntu image and that it doesn't do any harm on Ubuntu (besides the small lag).


spydon (approved):
Lgtm, unawaited sounds good

PR #638, opened 2024-04-12
feat(snap): remove tmp plug

matthew-hagemann:


d-loose (approved):
LGTM :+1:

Context: reverts #464, which is no longer needed after #548

PR #640, opened 2024-04-12
fix(bootstrap): map subiquity sections to pages in autoinstall

d-loose:
This simplifies the page setup logic:
* Get list of interactive sections from subiquity and map them to pages in the installer. If the result isn't empty, populate the wizard with those pages and ignore any `visible: false` from the config.
* If subiquity didn't provide any interactive sections, show all pages provided in the whitelabel.yaml, respecting `visible: false`.

I've also added a `required` property to the `InstallationStep` to simplify the logic in the wizard's `predicate` and keep everything in one place. Required pages are `loading`, `error`, `confirm`, `install` (for obvious reasons) and also `rst` and `secureBoot`, which weren't included before.

Fixes [autoinstall: interactive-sections: "*" causes the UI to hang](https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2060862)


spydon (approved):
Very nice, good improvement!

PR #641, opened 2024-04-12
fix(bootstrap): play welcome sound when finished loading

d-loose:
Fixes https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2061019


spydon (approved):
Makes sense

PR #650, opened 2024-04-16
fix(reset-tools): fix building factory-reset-tools snap

medicalwei:
This attempts to fix problems building factory-reset-tools as a snap, by:

* remove flutter library dependencies for -cli tools
* re-configure snapcraft.yaml to accommodate new directory structure, or adjust how we put files, so the snap package is buildable


packages/factory_reset_tools/lib/pages/creation_progress_page.dart
```diff
@@ -8,6 +8,19 @@ import 'package:ubuntu_wizard/ubuntu_wizard.dart';
 import 'package:yaru/foundation.dart';
 import 'package:yaru/widgets.dart';
 
+String displayCreationStatusName(
```
spydon, line 11: This can be done as an `extension` of the enum instead, then none of the other code has to change.

spydon (approved):
Lgtm, just one small comment

packages/factory_reset_tools/lib/pages/creation_progress_page.dart
medicalwei: Thanks, I followed your recommendation to use `extension on` instead of a standalone function.

spydon (approved):
Lgtm!

PR #653, opened 2024-04-16
chore: Generate l10n on weblate PRs

spydon:
Currently we have to do this manually on the weblate branches


d-loose (approved):
Awesome, thanks!

PR #656, opened 2024-04-16
Telemetry service

matthew-hagemann:
Adds a telemetry backend that calls ubuntu-report via its exported package.
Exposes the `Collect`, `CollectAndSend` and `SendDecline` methods


provd/internal/services/telemetry/telemetry.go
```diff
@@ -0,0 +1,119 @@
...
+	"strings"
+
+	pb "github.com/canonical/ubuntu-desktop-provision/provd/protos"
+	"github.com/ubuntu/ubuntu-report/pkg/sysmetrics"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// Option is a functional options for tests.
```
d-loose, line 17: Nittiest of nitpicks:

```suggestion
// Option is a functional option for tests.
```

d-loose (approved):
Thanks, LGTM :+1:

provd/internal/services/telemetry/telemetry.go
matthew-hagemann: Not at all! Good catch :smile:

PR #657, opened 2024-04-16
Init snap store updates

matthew-hagemann:
I committed these changes to the `init/ci` branch's `snapcraft.yaml`, but after they didn't reflect in the store request, I realize I should have updated this `snapcraft.yaml` instead :sweat_smile: 

Adds these commits as well as some style changes requested by the store team:
#638
#635
https://forum.snapcraft.io/t/request-for-system-files-interface-multiple-plugs-for-ubuntu-desktop-init/39717


d-loose (approved):
LGTM :+1:

PR #659, opened 2024-04-16
fix(bootstrap): ensure 'refresh' is only called once

d-loose:
Successive requests to `/refresh` cause subiquity some [troubles](https://bugs.launchpad.net/subiquity/+bug/2061756). Unfortunately the way the state of the "Refresh" button is handled allows sending multiple requests if the user double-clicks the button, as the UI state is only updated when the request completes. To fix this, I'm saving the refresh ID in between calls and make sure no additional request is made if another one is in progress already.


spydon (approved):
Lgtm, I guess the already existing future could have been bound to the variable instead too, but if there is a `refreshId` we can just as well use that.

PR #662, opened 2024-04-16
refactor(init): simplify ubuntu pro flow

d-loose:
* Removes unneeded `UbuntuProOnboardingModel`
* Adds a `skipPro` property to `UbuntuProModel` instead
* Use `skipPro` property to skip pro and success page instead of jumping to the privacy page


packages/ubuntu_init/test/init_wizard_test.dart
```diff
@@ -174,16 +176,11 @@ void main() {
       InitStep.locale.route,
       InitStep.keyboard.route,
       InitStep.identity.route,
-      InitStep.ubuntuProOnboarding.route,
-      InitStep.ubuntuPro.route,
-      InitStep.ubuntuProSuccess.route,
```
d-loose, line 179: I've made the 'init' test run through all steps instead, so we don't need those here.

packages/ubuntu_init/integration_test/screenshot_test.dart
```diff
@@ -106,6 +106,16 @@ Future<void> main() async {
...
+    await tester.jumpToPage(InitStep.ubuntuProOnboarding.route);
+    await tester.pumpAndSettle();
+
+    final context = tester.element(find.byType(UbuntuProOnboardingPage));
+    final l10n = UbuntuProLocalizations.of(context);
+
+    final option = find.text(l10n.ubuntuProOnBoardingEnableUbuntuPro);
+    await tester.tap(option);
+    await tester.pumpAndSettle();
+
```
d-loose, line 118: We need to select the correct option first, otherwise jumping to the page won't work because its `load` method will return `false.

d-loose:
@BLKKKBVSIK not sure why the PR workflow isn't working here. I ran the CI on this branch [here](https://github.com/d-loose/ubuntu-desktop-provision/actions/runs/8708776547) though

BLKKKBVSIK (approved):

PR #663, opened 2024-04-16
fix: Use `HorizontalPage` for `BitlockerPage`

spydon:
image


packages/ubuntu_bootstrap/lib/installer/installation_step.dart
```diff
@@ -20,6 +20,8 @@ enum InstallationStep with RouteName {
   sourceSelection(SourceSelectionPage.new, allowedToHide: true),
   codecsAndDrivers(CodecsAndDriversPage.new),
   notEnoughDiskSpace(NotEnoughDiskSpacePage.new, discreteStep: false),
+  bitlocker(BitLockerPage.new,
+      wizardStep: false, discreteStep: false, required: true),
```
d-loose, line 24: Is it necessary to move the `BitLockerPage` from the `StorageStep`s into the `InstallationStep`s? I find it a bit confusing, since it's still part of the `StorageWizard`.

spydon: That's true, I had some for of plan here that I didn't execute, but I reverted it now!

d-loose (approved):
LGTM :+1:

PR #683, opened 2024-04-18
fix: Tab should focus the next button when in Locale/Keymap list

spydon:


d-loose (approved):
Great, that's definitely an improvement :+1:

(Just tried navigating the entire installer with the keyboard and it's still a major pain unfortunately)

spydon:
> Great, that's definitely an improvement 👍
> 
> (Just tried navigating the entire installer with the keyboard and it's still a major pain unfortunately)

Indeed, this is just a first step

PR #690, opened 2024-04-19
fix: Add possiblity to tab out of list on wifi page

spydon:
This also refactors the wifi page to use providers groups instead of the static providers that it was overriding through the providerscope previously.


d-loose (approved):
LGTM :+1:

PR #694, opened 2024-04-22
ci: Enforce lockfiles for bootstrap and init

spydon:


spydon:
This needs https://github.com/invertase/melos/issues/703 before it will work

spydon:
Melos 6.0.0 is now released so this works like it should

d-loose (approved):
The console output of melos 6 also looks much nicer!

PR #697, opened 2024-04-22
feat: Use `langcode_countrycode` format for EULA files

### LINKED ISSUE #692: EULA file resolution ordering
Yao Wei has surfaced a need to change how we prioritize which EULA to surface: 

https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2063091

BLKKKBVSIK:
In this PR:

- Adding a fallback for locales with multiples language codes.

fix #692


packages/ubuntu_provision/lib/src/eula/eula_page.dart
```diff
@@ -22,12 +42,15 @@ class _EULAPageState extends ConsumerState<EULAPage> {
...
   Widget build(BuildContext context) {
-    final localeModel = ref.read(localeModelProvider);
     final lang = EULALocalizations.of(context);
-    final eulaFile = File(
-        '/usr/share/desktop-provision/eula/EULA_${localeModel.selectedLocale?.languageCode}.pdf');
-
-    Logger('eula').debug('EULA file: ${eulaFile.path}');
+    final eulaFile = ref.watch(eulaPageProvider).when(
+        data: (eulaFile) => eulaFile.path,
+        loading: () => File('/usr/share/desktop-provision/eula/EULA.pdf').path,
```
spydon, line 48: This will look strange, since it will quickly flip between the files, it's better to have a loader here instead.

packages/ubuntu_provision/lib/src/eula/eula_page.dart
```diff
@@ -1,15 +1,35 @@
...
+  final fileList = Directory('/usr/share/desktop-provision/eula').listSync();
+  final exp = RegExp(r'EULA_([a-zA-Z0-9\-]+)\.');
+  final supportedLocales = fileList
+      .map((entry) => basename(entry.path))
+      .where(exp.hasMatch)
+      .map((entry) => Locale(exp.firstMatch(entry)!.group(1)!));
+
+  final locale = basicLocaleListResolution([selectedLocale], supportedLocales);
+  final eulaFile = File(
+      '/usr/share/desktop-provision/eula/EULA_${locale.toLanguageTag()}.pdf');
```
spydon, line 26: This results in a different format than for the slides, I would say it is better to be consistent here and use the same format. We should also specify the format better in the customization guide.

spydon (approved):
Lgtm!

PR #703, opened 2024-04-23
fix(bootstrap): use udisks2-inhibit.service

dbungert:
udisks2-inhibit is being moved in the live environment to a service. Interact with that instead, instead of hardcoding the rules.

Depends on https://code.launchpad.net/~dbungert/livecd-rootfs/+git/livecd-rootfs/+merge/464842 is merged.

LP: #2063192

### DIFF
--- a/ci/snap/bootstrap/local/launcher
+++ b/ci/snap/bootstrap/local/launcher
@@ -41,10 +41,6 @@ fi
 [ "$needs_update" = true ] && echo "SNAP_DESKTOP_LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
 # Disable storage automounting (/usr/lib/udisks2/udisks2-inhibit)
-sudo mkdir -p /run/udev/rules.d
-sudo sh -c 'echo "SUBSYSTEM==\"block\", ENV{UDISKS_IGNORE}=\"1\"" > /run/udev/rules.d/90-udisks-inhibit.rules'
-trap "sudo rm -f /run/udev/rules.d/90-udisks-inhibit.rules; sudo udevadm control --reload; sudo udevadm trigger --subsystem-match=block" EXIT HUP INT QUIT ILL ABRT FPE KILL SEGV PIPE ALRM TERM BUS
-sudo udevadm control --reload
-sudo udevadm trigger --subsystem-match=block
-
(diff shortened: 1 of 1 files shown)


dbungert:
Draft because I haven't tested it yet.

ci/snap/bootstrap/local/launcher
```diff
@@ -41,10 +41,6 @@ fi
 [ "$needs_update" = true ] && echo "SNAP_DESKTOP_LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
 # Disable storage automounting (/usr/lib/udisks2/udisks2-inhibit)
-sudo mkdir -p /run/udev/rules.d
-sudo sh -c 'echo "SUBSYSTEM==\"block\", ENV{UDISKS_IGNORE}=\"1\"" > /run/udev/rules.d/90-udisks-inhibit.rules'
-trap "sudo rm -f /run/udev/rules.d/90-udisks-inhibit.rules; sudo udevadm control --reload; sudo udevadm trigger --subsystem-match=block" EXIT HUP INT QUIT ILL ABRT FPE KILL SEGV PIPE ALRM TERM BUS
-sudo udevadm control --reload
-sudo udevadm trigger --subsystem-match=block
-
+trap "sudo systemctl stop udisks2-inhibit.service" EXIT HUP INT QUIT ILL ABRT FPE KILL SEGV PIPE ALRM TERM BUS
```
ogayot, line 44: I think I'd rather remove this stuff completely.

I'm worried that restarting the installer for one reason or another would end up reactivating and triggerring automatic mounts.

dbungert: I think that's intended behavior by people trying to use the live-environment for non-installer things, but yes that would be more robust from the installer perspective to just not automount them.

d-loose: Agreed, we rely on closing and manually restarting the installer to update it. That would mean there'd be no automounting in the live session at all, though - even if the user chooses 'Try Ubuntu'.

dbungert: Marking resolved as I think removing it is not an option.

d-loose (approved):
Thanks, feel free to merge this once the corresponding livecd-rootfs change is in

Issue #710, opened 2024-04-24
Country codes now required on all language files


matthew-hagemann:
Yao Wei reported that the fix to resolve EULA files correctly has had the unintended consequence of forcing all EULA files to need country codes which they had to add manually:

EULA_de.pdf -> EULA_de_DE.pdf

spydon:
That is intended and as specified in the customization docs, we don't want to support multiple formats.

medicalwei:
Ah, just noticed that we have discussions here.

I'd like to confirm internally about whether we can rename the files, especially considering this is a legal document.

medicalwei:
As for the document https://docs.google.com/document/d/10R0YOj4e8BTv6XPw9OE_y7GDy72xPqA5XP5lu0M7VbE/edit

The requirements of the language code format is not mentioned explicitly, and I did not realize that this requirement also applies to EULA files.  I think the document may need to document what language codes ubuntu-desktop-provision can use.  If that's clear, I do not have much concern.

spydon:
> As for the document https://docs.google.com/document/d/10R0YOj4e8BTv6XPw9OE_y7GDy72xPqA5XP5lu0M7VbE/edit
> 
> The requirements of the language code format is not mentioned explicitly, and I did not realize that this requirement also applies to EULA files.  I think the document may need to tell what language codes ubuntu-desktop-provision can use for their assets.  If that's clear, I do not have much concern.

It does say "The language code format is the same as is used for slides, for example: EULA_en_US.pdf", but I can clarify it further.

spydon:
@medicalwei I added a new section in the customization guide further detailing the language code format that is used:
https://docs.google.com/document/d/10R0YOj4e8BTv6XPw9OE_y7GDy72xPqA5XP5lu0M7VbE/edit#heading=h.2463op6ywahp

PR #715, opened 2024-04-25
fix(eula): find locale with flutter resolution algorithm

medicalwei:
This attempts to fix EULA language by finding available EULA languages first, and match them against preferred language using a matching function in Flutter.

If a test function is needed, we'd like to know how we could implement it.


spydon:
As noted in #710, I don't think that we should support multiple language code formats, it will also be inconsistent with the slides. How come the EULA files can't follow one format, as specified in the customization docs?

medicalwei:
Close this PR although we may have to revisit this for 25.04

Issue #717, opened 2024-04-25
Remove the usage of `SafeNotifierProvider`


spydon:
Instead of using the `SafeNotifierProvider` we should refactor those providers to use built-in providers in Riverpod, like `FutureProvider`, `StreamProvider` and `StateProvider`.

PR #718, opened 2024-04-25
refactor: Use built-in providers for installer model

spydon:
Part of #717


packages/ubuntu_bootstrap/test/confirm/confirm_page_test.dart
```diff
@@ -78,9 +78,13 @@ final testDisks = <Disk>[
 
 void main() {
   setUp(() {
+    final mockInstallerService = MockInstallerService();
+    when(mockInstallerService.monitorStatus()).thenAnswer(
+      (_) => Stream.value(fakeApplicationStatus(ApplicationState.RUNNING)),
+    );
```
d-loose, line 84: Would it maybe be better to override the new `applicationStatusProvider` in `buildConfirmPage`, instead of adding a mock response to the service?

spydon: Yeah, it's either or I guess. Like this the code goes through the real provider too so one could argue that it gives slightly more code coverage, in this case I don't think it matters, so possibly we should override the provider since that is what we're doing in many other places.

d-loose: Ah I see, good point. Let's talk about it sometime next week and agree on a consistent approach so that testing in this code base becomes less of a nightmare :)

ogres:
@spydon Is this PR still valid?

spydon:
@ogres it is, since I don't have write rights in this repository I re-implemented it in #1513

PR #719, opened 2024-04-25
fix(init): homedir cannot be /run, use tmp

matthew-hagemann:


spydon (approved):

ogres:
@matthew-hagemann Is this still valid?

matthew-hagemann:
@ogres Yes, but only if we want to continue with the `ubuntu-desktop-init` project, which we put to the backlog a few years ago.

PR #721, opened 2024-04-26
feat(reset-tools): change the boot option "type" to "key" string

medicalwei:
This feature was intended to use string type key, as it is configurable from the file in /usr/share/desktop-provision/reset.yaml to add some customizable configurations for OEM. Using enum type breaks this design but I am wondering whether we have other concerns from Desktop Team.


spydon (commented):
I'd prefer is the string was taken from the enum when it is needed, if we just have a pure string we lose some type safety. So just do `ResetOptionType.factoryReset.value` etc where the string is needed.

medicalwei:
I see, although just using the string from enum would not address my concern, I will fix what is visible from me.

PR #722, opened 2024-04-26
fix(reset-tools): fix listing options in reboot in cli

medicalwei:
When running `factory-reset-tools.cli reboot`, it shows the messages below:

```
List of available options:

ResetOptionType.factoryReset: Restore Ubuntu to factory state
  This option will restore Ubuntu to factory default, removing all files stored in this system during the process.

ResetOptionType.fwSetup: UEFI Firmware Settings
  Restart into UEFI Firmware (BIOS) Settings menu.
```

Of which `ResetOptionType.factoryReset` is the enum string, and this should be the enum value `factory-reset` so when entering `factory-reset-tools.cli reboot factory-reset` it should start the reset.


spydon (approved):
Lgtm!

spydon:
I'm fixing the mocks issue, I'll merge this once that is in

PR #724, opened 2024-04-26
fix(provd): unlock login keyring on startup

### LINKED ISSUE #711: Connect to wifi doesn't have password input popup
When selecting Wi-Fi for connection, the UI gets stuck and does not ask for a password.

https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2063313

d-loose:
This runs `gnome-keyring-daemon --unlock` with a temporary password before launching the provd daemon, so that network manager doesn't complain about missing agents. By default the network credentials will be stored in `/etc/netplan` instead of the user's keyring, though. To fix [lp:2063313](https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2063313) we don't need to worry about copying the keyring to the new user's directory and updating the password. I'd open a separate issue for that. 

Fix #711

### DIFF
--- a/provd/cmd/provd/daemon/daemon.go
+++ b/provd/cmd/provd/daemon/daemon.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"os/exec"
 	"runtime"
 
 	"github.com/canonical/ubuntu-desktop-provision/provd/internal/daemon"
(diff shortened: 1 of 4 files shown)


spydon (approved):

d-loose:
Locally build a deb and tested that this works. I'll merge it now so that a deb artifact is available for the OEM team to test it.

PR #736, opened 2024-05-07
chore(bootstrap): add bugtracker url to snapcraft.yaml

d-loose:
When running `ubuntu-bug ubuntu-desktop-bootstrap` the following error is shown:

image

I suspect we need to add the corresponding URL to the snapcraft.yaml

Will make the link on the error page invoke `ubuntu-bug` in a separate PR, once this works :)


spydon (approved):

seb128 (approved):
Thanks

PR #760, opened 2024-06-05
feat(bootstrap): support local file:// URIs on autoinstall page

### LINKED ISSUE #754: Autoinstall page: support local paths
In addition to remote URLs we've been asked to support paths in the local file system for selecting an autoinstall.yaml via the UI. A typical use case would be: downloading an autoinstall.yaml via a browser in the live session after going through an authentication flow.

We agreed that no file picker dialog is needed and supporting `file:///path/to/autoinstall.yaml` is sufficient.

d-loose:
Adds support for local file paths following the file URI scheme on the autoinstall page (e.g. `file:///home/ubuntu/foo.yaml`.

Successfully tested this with the 24.04 iso.

image

@anasereijo do we need to update the copy here? As far as I understand this feature is mainly needed for internal purposes.

Fix #754
UDENG-3038

### DIFF
--- a/packages/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_model.dart
+++ b/packages/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_model.dart
@@ -75,20 +75,26 @@ class AutoinstallModel extends SafeChangeNotifier {
 
   Future<void> _fetch() async {
     final uri = Uri.parse(url);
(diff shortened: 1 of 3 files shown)


spydon (approved):
Lgtm!

anasereijo:
@d-loose to see if I understand correctly, we now support the user to either enter a URL or a local file path, right?

The text now reads 'Enter ... URL' - Maybe we should change the text to say 'URL or local file path'. wdyt?

PR #762, opened 2024-06-06
chore: Don't override yaru in factory_reset_tools

spydon:
Also added the lock file and defined the flutter version for the factory_reset_tools


d-loose (approved):
LGTM :+1:

PR #765, opened 2024-06-11
generator: accept multiple .py files as input

ogayot:
The types.py file in the Subiquity tree is getting difficult to maintain and it would be ideal to split it into multiple files going forward.

Ensure that the generator script can parse content from multiple files.

NOTE: Previously the syntax `./generator.py file1 file2` would parse "file1" and output to "file2". Now, this syntax would parse "file1" and "file2" ; and would output to the standard output. The generator will look for specific file extensions when reading (i.e., *.py) and writing (i.e., *.dart) so there should be no risk of overwriting important files if using the generator incorrectly.

### DIFF
--- a/packages/subiquity_client/generator/Makefile
+++ b/packages/subiquity_client/generator/Makefile
@@ -2,7 +2,7 @@ types_py = ../subiquity/subiquity/common/types.py
 types_dart = ../lib/src/types.dart
 
 generate:
-	python3 generator.py $(types_py) $(types_dart)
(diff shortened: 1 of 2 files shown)


d-loose (approved):
Thanks a lot for taking care of that in advance!

LGTM :+1:

d-loose:
Ah, a tiny nitpick: could you please sign your commit? :)

ogayot:
> Ah, a tiny nitpick: could you please sign your commit? :)

Sure, done! :+1:

PR #770, opened 2024-06-14
feat: run `ubuntu-bug` when error page is shown

d-loose:
There are some issues with the classic snap environment when trying to run ubuntu-bug:
```
ubuntu@ubuntu:~$ snap run --shell ubuntu-desktop-bootstrap
ubuntu@ubuntu:~$ ubuntu-bug ubuntu-desktop-bootstrap
Cannot start: /lib/x86_64-linux-gnu/libgirepository-1.0.so.1: undefined symbol: g_once_init_leave_pointer
```

After adding libgirepository-1.0.so.1 to the stage-packages I get the following instead:
```
ubuntu@ubuntu:~$ snap run --shell ubuntu-desktop-bootstrap
ubuntu@ubuntu:~$ ubuntu-bug ubuntu-desktop-bootstrap
Cannot start: /usr/lib/python3/dist-packages/gi/_gi.cpython-312-x86_64-linux-gnu.so: undefined symbol: g_assertion_message_cmpint
```

I’ve tried staging python3-gi as well, but couldn’t resolve it yet.

Not sure how deep that rabbit hole goes.

UDENG-3172


packages/ubuntu_provision/lib/src/locale/locale_page.dart
```diff
@@ -14,6 +14,7 @@ class LocalePage extends ConsumerWidget with ProvisioningPage {
   Future<bool> load(BuildContext context, WidgetRef ref) async {
     final model = ref.read(localeModelProvider);
     await model.init();
+    throw Exception('Test Error');
```
d-loose, line 17: For testing purposes this exception is thrown when the language selection page is loaded.

PR #777, opened 2024-06-25
 Allow to hide pages don't needed in core desktop

sergio-costas:
Core desktop doesn't require the UbuntuPro page. Also, at the moment, the Accessibility doesn't work. So those two pages must be allowed to be hiden.


d-loose (approved):
LGTM :+1:

spydon (approved):
Lgtm!

sergio-costas:
Thanks!

PR #783, opened 2024-06-27
feat(bootstrap): include subiquity late commands fix

d-loose:
This includes the subiquity [fix](https://github.com/canonical/subiquity/pull/2013) for [lp:2065958](https://bugs.launchpad.net/subiquity/+bug/2065958) in the desktop installer.

I've added a test case that ensures the `LATE_COMMANDS` status is treated correctly, i.e. leaves the installation page in the "installing" state.

Supersedes #781
UDENG-3382


matthew-hagemann (approved):
LGTM

PR #784, opened 2024-07-01
provd: Removing pro and chown services (sprovd)

matthew-hagemann:
This change drops `sprovd` for security concerns. With it I have removed the `chown` and `pro` services that relied on it from the go backend, and the `chown` service from the Flutter frontend.

- The `chown` work will be made redundant by future `systemd` work for provd
- The `pro` backend will be moved to stage 5 / Flutter in the future.

While this does temporary pose an issue for the a11y and pro flutter pages, we can disable them thanks to #777 for the time being.

Closes Udeng 3455


spydon:
Now this is just a draft, but the PR titles should be a bit more explanatory I think :smile:

d-loose (commented):
As this renders the ubuntu pro page in ubuntu init non-functional, should we also remove the UI and service for that?

matthew-hagemann:
> As this renders the ubuntu pro page in ubuntu init non-functional, should we also remove the UI and service for that?

We do ultimately want to write a dart backend for pro and keep the page. My thinking was to just leave it for now until we get back to that, as it can be disabled, and only the core desktop team is using it? I'm happy to remove more than I have though if you think that would be better.

d-loose (approved):
I see, that's fine with me then :)

PR #789, opened 2024-07-08
feat(provision): show message to run ubuntu-bug on error page

d-loose:
Since automatically running `ubuntu-bug` currently doesn't work due to classic snap confinement issues, @seb128 suggested displaying a corresponding message on the error page instead.

Screenshot from 2024-07-08 12-05-22

UDENG-3172


d-loose:
Added `sudo` to the message, as the live session user can't access the subiquity logs, (but they're able to invoke `sudo` without entering a password).

seb128:
Thanks, that's looking good to me!

PR #796, opened 2024-07-30
fix(reset-tools/reset_media.dart): fix exceptions on unmounting

medicalwei:
We received a report that, if using a disk without partition table:

```
sudo dd if=/dev/zero of=/dev/sda bs=512 count=1
sudo mkfs.fat /dev/sda
```

is it is possible to trigger error that USB media remains unformatted.

However, after my investigation I also found that partition unmounted error is not handled properly.  This merge request handles both `org.freedesktop.UDisks2.Error.NotMounted` error and treat it as nothing happened, as well as properly unmounting USB media without partition table.

I am waiting for colleague's response before actual merge.

### DIFF
--- a/packages/factory_reset_tools/lib/dbus/reset_media.dart
+++ b/packages/factory_reset_tools/lib/dbus/reset_media.dart
@@ -24,7 +24,7 @@ class Drive {
   Drive(this.object);
   late final DBusRemoteObject object;
 
-  Future<Partition> format() async {
+  Future<Filesystem> format() async {
     final formatTableParams = [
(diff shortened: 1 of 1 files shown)


medicalwei:
I got response from @cyruslien that this PR fixed his use case.

d-loose (approved):
LGTM :+1:

@spydon is off this week, not sure if you want/need to wait for his approval as well.

medicalwei:
This change does not change the UI (flutter) side of things, and I think it is okay for him to comment on the code and improve it later.

PR #799, opened 2024-08-20
Setup sphinx

s-makin:
This PR will prepare the docs to be added to RTD. It is composed of two commits:
1) To add Sphinx files from the Starter Pack
2) To add the docs pages and link all together

- Some content is not required immediately - those lines have been commented out
and can be included once ready to move to Diataxis.
- Diataxis structural files can be moved from `./diataxis_docs/` to `./docs` at that point, and should build fine (although the text on each file can be adjusted to fit your needs).
- Some links cannot be populated in custom_conf.py until after the docs are imported
to RTD, so some small adjustments will need to be made afterwards.
- Check: `ogp_site_url`, `product_page`, `discourse`, `mattermost`.
- `slug` can be set once the URL is set on RTD to documentation.ubuntu.com/<something>.


matthew-hagemann (approved):
Awesome, thank you Sally!

PR #803, opened 2024-08-28
feat(reset-tools): Add permission check based on polkit

medicalwei:
This feature adds polkit authorization before performing factory reset, both in GUI and in CLI

Need to request for an additional `polkit` autoconnect: https://forum.snapcraft.io/t/factory-reset-tools-polkit-autoconnect-request/41632

While implementing polkit interface for GUI applications seems to be straightforward by adding authorization on the server side, doing it in the CLI client is quite a problem.

First, it is necessary to spawn `pkttyagent` by the user program to show the text prompt, but to wait for the `pkttyagent` to be registered onto `polkitd`, `pkttyagent` uses a `--notify` channel which passes a file descriptor that it should close, which, referring to [fwupd implementation](https://github.com/fwupd/fwupd/blob/main/src/fu-polkit-agent.c), uses a pipe to implement.  However, Dart does not have a low level file/pipe mechanism available, therefore we have to use `dart:ffi` and a C helper to implement such feature.

Second, snapd seems not allow using `pkttyagent` without using `--fallback` to properly show text prompt. The discussion is in Snap Forum: https://forum.snapcraft.io/t/polkit-interface-and-pkttyagent-in-snap/41616

Internal Reference: SOMERVILLE-600

### DIFF
--- a/ci/snap/reset-tools/snapcraft.yaml
+++ b/ci/snap/reset-tools/snapcraft.yaml
@@ -38,6 +38,7 @@ apps:
       - hardware-observe
       - mount-observe
       - dbus-system-com-canonical-oem-factory-reset-tools-client
+      - polkit
   dbus-daemon:
     command: bin/factory-reset-tools-cli dbus
     daemon: simple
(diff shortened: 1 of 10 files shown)


spydon (approved):
Lgtm!

PR #809, opened 2024-09-05
feat: add 20th anniversary ubuntu logo

d-loose:
UDENG-4412


spydon (commented):
Looks like the bounding boxes are a bit off? :thinking:

d-loose:
Yes, noticed that too - double-checking with the design team

d-loose:
I'd propose merging it now to test the new pipeline and get a 24.10 build with the new mascot/logos in place. We can still land fixes later if we need to :)

spydon (approved):

PR #810, opened 2024-09-05
feat: release new version with updated mascot and logo

d-loose:
Includes the new mascot and logos from #809 as well as the latest subiquity fixes from the `ubuntu/noble` branch (#807) (I'll open a follow-up PR tomorrow to track `ubuntu/main` and pull in the latest subiquity changes here).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -49,7 +49,7 @@ parts:
     after: [flutter-git]
     plugin: nil
     source: .
-    source-commit: &commit-ref 298edf271695df828a169f153143da18712739c3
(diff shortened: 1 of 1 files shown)


d-loose:
I can't figure out why the script in the CI isn't working now, but I manually checked that the dependencies match. I'll have a look at the script tomorrow

d-loose:
Looking at the [docs of the checkout action](https://github.com/actions/checkout?tab=readme-ov-file#usage) it seems like we need to set `fetch-depth` to `0`, so it checks out the entire commit history. Otherwise we can't access files from other points in history. Will test that tomorrow.

PR #769, opened 2024-06-13
generator: update to match Subiquity moved API types

ogayot:
Subiquity is moving API types away from subiquity/common/types.py into subiquity/common/types/{__init__,storage}.py. Ensure the makefile follows the new architecture.

Marking as a draft for now. This will be needed once https://github.com/canonical/subiquity/pull/2010 lands in udi.

PR #811, opened 2024-09-06
feat(bootstrap): track subiquity's main branch

d-loose:
This sets up the main development branch in this repo to track subiquity's `main` branch. I've included @ogayot's change #769 for the type generator and used it to update the Dart classes. I also removed some remaining parts of the decommissioned WSL setup (see internal spec WS030), since the the corresponding types have been removed from subiquity.

For additional 24.04.x bug fixes we will use the new `ubuntu/24.04`, which still tracks subiquity's `ubuntu/noble` branch.

I'm also removing the remaining snapcraft files for ubuntu-desktop-bootstrap here.
We're migrating to the new setup (internal specs UD064 and UD068) with the snapcraft files now located on dedicated branches (`snap/ubuntu-destkop-bootstrap/main` in this case).
I'll follow up with a proper documentation update next week.


d-loose:
I've now built a snap, injected this into the current daily (20240906) and done a successful install.

spydon (approved):
Lgtm

PR #815, opened 2024-09-09
docs: adding how to add language packs in autoinstall.yaml

matthew-hagemann:
Adding a section on how to install additional language packs to appear in GIS via running late-commands in the `autoinstall.yaml`

This is doubled up as its applicable to two flow's

### DIFF
--- a/docs/oem-provisioning-24_04_1.md
+++ b/docs/oem-provisioning-24_04_1.md
@@ -135,8 +135,7 @@ autoinstall:
       curtin in-target --target=/target -- bash -c '
         # Update the package list
         apt-get update
-        # Installing languages not in seed
-        # Note that firefox and thunderbolt are snap packages, so these are not added here
+        # Installing extra languages
         for lang in "fr" "de"; do
              for pkg in $(check-language-support --show-installed -l "$lang"); do
                   # gimp: not installed by default
@@ -237,8 +236,7 @@ autoinstall:
       curtin in-target --target=/target -- bash -c '
         # Update the package list
         apt-get update
-        # Installing languages not in seed
-        # Note that firefox and thunderbolt are snap packages, so these are not added here
+        # Installing extra languages
         for lang in "fr" "de"; do
              for pkg in $(check-language-support --show-installed -l "$lang"); do
                   # gimp: not installed by default
--- a/docs/oem-provisioning-24_04_1.md
+++ b/docs/oem-provisioning-24_04_1.md
@@ -119,11 +119,12 @@ touch meta-data
 ```
 #### Adding additional Language packs
 
-The default Ubuntu ISO only has the English language pack seeded. To have additional languages
-available in Gnome Initial Setup, you will need to install additional language packs via the
-`autoinstall.yaml`
+The default Ubuntu ISO seeds multiple languages, but only carries one to the target system. To have
+additional languages available in Gnome Initial Setup, you will need to install additional language
+packs via the `autoinstall.yaml` onto the target system
 
-To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands) in the `autoinstall.yaml:
+To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands)
+in the `autoinstall.yaml:
 
 ```yaml
 #cloud-config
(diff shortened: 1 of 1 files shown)


docs/oem-provisioning-24_04_1.md
```diff
@@ -117,7 +117,42 @@ autoinstall:
 EOF
 touch meta-data
 ```
+#### Adding additional Language packs
 
+The default Ubuntu ISO only has the English language pack seeded. To have additional languages available in Gnome Initial Setup, you will need to install additional language packs via the `autoinstall.yaml`
```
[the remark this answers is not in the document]
matthew-hagemann: oh gosh I do it every time :joy: ty!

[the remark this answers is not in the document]
matthew-hagemann: Should I handle the links? They themselves are > 100 characters?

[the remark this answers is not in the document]
spydon: No, links are fine to overflow, since you're not reading them, so the text still reads fine in non-wrapping environments. :)

docs/oem-provisioning-24_04_1.md
```diff
@@ -117,7 +117,44 @@ autoinstall:
...
+                # mozc-utils-gui: is in universe
+                if [ "${pkg%%-*}" = gimp ] || [ "${pkg%%-*}" = fcitx5 ] || [ "$pkg" = "mozc-utils-gui" ]; then
+                    continue
+                fi
+                apt-get install -y "$pkg"
+            done
+            for pkg in $(check-language-support --show-installed -l "$lang" -p "ibus,gnome-user-docs,gvfs"); do
+                apt-get install -y "$pkg"
+            done
+            for pkg in $(check-language-support --show-installed -l "$lang" -p "libreoffice-common"); do
```
seb128, line 153: We don't install libreoffice by default (unless you change the option during the installation to install extra software) so that should probably be conditional on whether libreoffice-common is installed?

docs/oem-provisioning-24_04_1.md
```diff
@@ -117,7 +117,44 @@ autoinstall:
...
+            for pkg in $(check-language-support --show-installed -l "$lang"); do
+                # gimp: not installed by default
+                # fcitx5: not used by ubuntu as gnome is in use
+                # mozc-utils-gui: is in universe
+                if [ "${pkg%%-*}" = gimp ] || [ "${pkg%%-*}" = fcitx5 ] || [ "$pkg" = "mozc-utils-gui" ]; then
+                    continue
+                fi
+                apt-get install -y "$pkg"
+            done
+            for pkg in $(check-language-support --show-installed -l "$lang" -p "ibus,gnome-user-docs,gvfs"); do
```
seb128, line 150: why do we special case those? would be nicer if the reason was documented

seb128 (changes requested):
Thanks for the work there, I've added a few ask for things to change or fix

docs/oem-provisioning-24_04_1.md
matthew-hagemann: Ah, and additionally, don't we install the snap by default if you choose the normal install? Should I actually always skipping libre office related language support?

docs/oem-provisioning-24_04_1.md
matthew-hagemann: My mistake, it doesn't make sense in our context, the OEM team have a different script they run, not `apt-get install`. These will be picked up by the first for loop.

docs/oem-provisioning-24_04_1.md
```diff
@@ -117,7 +117,44 @@ autoinstall:
 EOF
 touch meta-data
 ```
+#### Adding additional Language packs
 
+The default Ubuntu ISO only has the English language pack seeded. To have additional languages
+available in Gnome Initial Setup, you will need to install additional language packs via the
+`autoinstall.yaml`
+
+To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands) in the `autoinstall.yaml:
```
spydon, line 126:
```suggestion
To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands)
in the `autoinstall.yaml:
```

docs/oem-provisioning-24_04_1.md
```diff
@@ -187,6 +224,45 @@ EOF
 touch meta-data
 ```
 
+#### Adding additional Language packs
+
+The default Ubuntu ISO only has the English language pack seeded. To have additional languages
+available in Gnome Initial Setup, you will need to install additional language packs via the
+`autoinstall.yaml`
+
+To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands) in the `autoinstall.yaml:
```
spydon, line 233:
```suggestion
To install additional language packs, you can make use of [late commands](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#late-commands)
in the `autoinstall.yaml:
```

matthew-hagemann:
Thanks for the review @seb128 :slightly_smiling_face:  I've simplified the conditions a little and documented them better. Skipping libreoffice language packs is not conditional based on if `libreoffice-core` is installed

spydon (approved):
Lgtm!

seb128 (approved):
Thanks, that looks fine to me now. I just have one extra nitpick about the comments inline but feel free to commit once that's changed

PR #822, opened 2024-09-17
feat: Inherit Subiquity's ability to configure block probing timeout

ogayot:
Port of https://github.com/canonical/subiquity/pull/2073 to ubuntu-desktop bootstrap.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+# source the management script
+. "$SNAP"/management-script
+
+
+handle_block_probing_timeout_config()
+{
+    block_probing_timeout="$(block_probing_timeout)"
+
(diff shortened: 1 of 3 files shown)


ogayot:
Looks like it's working now.

Supported commands are:

```
# Set the timeout back to the default value (120)
snap unset ubuntu-desktop-bootstrap block-probing-timeout

# Disable the timeout
snap set ubuntu-desktop-bootstrap block-probing-timeout=no
snap set ubuntu-desktop-bootstrap block-probing-timeout=null

# Set the timeout to a specific value
snap set ubuntu-desktop-bootstrap block-probing-timeout=240
```

d-loose (approved):
Thanks, I've successfully built a snap and tested that setting the timeout works :+1:

PR #828, opened 2024-09-19
ci: add snap workflow

d-loose:
Adds a snapcraft workflow to the CI that runs on pushes and pull requests to `snap/**` branches.
I'd also add this to the other `snap/**` branches.

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -1,0 +1,27 @@
+name: Snap
+
+on:
+  push:
+    branches:
+      - snap/**
+  pull_request:
+    branches:
+      - snap/**
+  workflow_dispatch:
+
+jobs:
+  snap:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: snapcore/action-build@v1
+        with:
+          snapcraft-channel: 7.x
+        id: snapcraft
+      - uses: actions/upload-artifact@v4
+        if: github.event_name == 'workflow_dispatch'
+        with:
+          name: 'snap'
+          path: ${{steps.snapcraft.outputs.snap}}


Chris-Peterson444 (approved):

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: snapcore/action-build@v1
+        with:
+          snapcraft-channel: 7.x
+        id: snapcraft
+      - uses: actions/upload-artifact@v4
+        if: github.event_name == 'workflow_dispatch'
```
matthew-hagemann, line 24: Do we want to make the upload of the snap run on a different condition to building the snap?

matthew-hagemann (changes requested):
Looks good! One quick question though:

.github/workflows/snap.yaml
d-loose: So far we've been uploading build artifacts only for manually triggered workflows in all our repos to avoid any potential issues with aritfact storage quota.
I think we could do an upload for every push to the snap branch though, since those don't happen too frequently (but still exclude pull requests) - what do you think?

matthew-hagemann: I guess now that they are on their own branch we can see if its okay? 

Otherwise I was more wondering are we building but not uploading on push as a test, and then triggering on a workflow the upload for when we want the output snap?

matthew-hagemann: Ah okay, then I think what you have makes total sense. Always validate with builds, when we want the output we ask for it. Sorry for going in a circle to get me following :slightly_smiling_face:

matthew-hagemann (approved):

.github/workflows/snap.yaml
d-loose: Not at all - very valid points! As discussed yesterday, it might be useful to retain build artifacts from workflows running on PRs and publish them to a short-lived snap branch for testing. But we'll need to setup the pipeline for that first

PR #834, opened 2024-09-23
chore: remove unused stage-packages

d-loose:
Same as #833 but for 24.04

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -144,9 +144,6 @@ parts:
       - python3.10-minimal
       - ssh-import-id
       - ubuntu-advantage-tools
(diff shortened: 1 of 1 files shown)


matthew-hagemann (changes requested):
Do we also need to remove `dctrl-tools` on line 117?

d-loose:
Thanks for double-checking! I should have commented on the difference in those two PRs - the dependency got removed in upstream subiquity [here](https://github.com/canonical/subiquity/commit/ac54d83eae088bf1962c27a061374da10c67a251)

matthew-hagemann (approved):
Ah okay, perfect. LGTM!

PR #835, opened 2024-09-23
fix: use UTF-8 encoding for password hashing

### LINKED ISSUE #831: FIX: ubuntu installer doesnt work with some specific chars
On Ubuntu 24.04, I attempted various password combinations, both with and without special characters, but the password I entered during installation didn't work on the installed OS.

I specifically tried using the character 'ù' but couldn’t log in after the installation.

To work around this installer bug, use a simple password during installation, and then change it to a more complex one afterward.

d-loose:
This fixes a bug where a password containing non-ASCII characters set during the installation would be rejected when booting the newly installed system.
As Dart [`String`](https://api.dart.dev/stable/3.5.3/dart-core/String-class.html)s consist of UTF-16 encoded code units, the SHA-512 crypt hash computed with [`crypt.dart`](https://pub.dev/packages/crypt) doesn't match with the one computed when the user logs in, where the encoding will [default to UTF-8](https://github.com/canonical/ubuntu-desktop-provision/blob/4a1c55e84cd0a3955c7e2b9d77b212a1632bf0f6/packages/ubuntu_provision/lib/src/locale/locale_model.dart#L71-L75).

Fix #831
UDENG-4710

### DIFF
--- a/packages/ubuntu_bootstrap/lib/services/identity_service.dart
+++ b/packages/ubuntu_bootstrap/lib/services/identity_service.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 import 'package:crypt/crypt.dart';
 import 'package:meta/meta.dart';
 import 'package:subiquity_client/subiquity_client.dart';
@@ -5,10 +7,15 @@ import 'package:ubuntu_bootstrap/services/post_install_service.dart';
 import 'package:ubuntu_provision/services.dart';
 
 class SubiquityIdentityService implements IdentityService {
-  const SubiquityIdentityService(this._subiquity, this._postInstall);
+  const SubiquityIdentityService(
+    this._subiquity,
+    this._postInstall, {
+    @visibleForTesting this.testSalt,
+  });
 
   final SubiquityClient _subiquity;
   final PostInstallService _postInstall;
+  final String? testSalt;
 
   /// The auto-login post-install config key for testing purposes.
   @visibleForTesting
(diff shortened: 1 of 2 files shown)


packages/ubuntu_bootstrap/lib/services/identity_service.dart
```diff
@@ -1,14 +1,18 @@
...
 import 'package:crypt/crypt.dart';
 import 'package:meta/meta.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_bootstrap/services/post_install_service.dart';
 import 'package:ubuntu_provision/services.dart';
 
 class SubiquityIdentityService implements IdentityService {
-  const SubiquityIdentityService(this._subiquity, this._postInstall);
+  const SubiquityIdentityService(this._subiquity, this._postInstall,
+      {@visibleForTesting this.testSalt});
```
spydon, line 11: Trailing comma

packages/ubuntu_bootstrap/lib/services/identity_service.dart
```diff
@@ -37,7 +41,10 @@ class SubiquityIdentityService implements IdentityService {
       IdentityData(
         realname: identity.realname,
         username: identity.username,
-        cryptedPassword: Crypt.sha512(identity.password).toString(),
+        cryptedPassword: Crypt.sha512(
+          String.fromCharCodes(utf8.encode(identity.password)),
```
spydon, line 45: Soo many people must have done this mistake, good catch!

spydon (approved):
Lgtm!

packages/ubuntu_bootstrap/lib/services/identity_service.dart
d-loose: Ah, we're still using `ubuntu_lints` 0.3.0 here.. thanks!

sminez (approved):

PR #858, opened 2024-11-06
feat: moving guided capabilities to a dedicated page

sminez:
This is primarily intended as a small change to the code base for myself and @matthew-hagemann to get a better understanding of the overall architecture and flow for updating things.

Off the back of this there are several changes we want to make to how the integration tests are written and managed as well as writing some documentation for contribution and maintenance of the project as a whole.

cc. @anasereijo If we want to do this then we probably want / need a new illustration for the new page rather than simply reusing the one from the storage page. There is also no rush / need to merge this any time soon but just so it is on your radar :slightly_smiling_face:


anasereijo:
@sminez could you please share screenshots of this so that I can review and help out with the layout/copy? thank you

sminez:
@anasereijo I can get you some screenshots of the current state but please be aware that the work on this branch is not going to be merged as it currently stands. @matthew-hagemann is looking to rework it this week in order to fit it in with the changes we've been making to the installer so far this cycle.

sminez:
Current UI state from this PR

PR #859, opened 2024-11-07
refactor: move tester navigation out of integration tests

matthew-hagemann:
A small refactor to move repeated navigation functions into their test functions for better test readability to help me get more comfortable with navigating the installer project :slightly_smiling_face:

### DIFF
--- a/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
+++ b/packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
@@ -43,8 +43,7 @@ Future<void> main() async {
       await tester.runApp(() => runInstallerApp([], theme: currentTheme));
       await tester.pumpAndSettle();
 
-      await tester.testLocalePage(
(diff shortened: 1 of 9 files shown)


packages/ubuntu_bootstrap/integration_test/screenshot_test.dart
```diff
@@ -43,9 +43,7 @@ Future<void> main() async {
       await tester.runApp(() => runInstallerApp([], theme: currentTheme));
       await tester.pumpAndSettle();
 
-      await tester.testLocalePage(
-        screenshot: '$currentThemeName/locale',
-      );
+      await tester.testLocalePage(screenshot: '$currentThemeName/locale');
```
matthew-hagemann, line 46: 1) For some reason, we don't require the navigation flag in certain bootstrap integration tests

d-loose (approved):
Thanks, that's a lot more legible! :+1:

PR #862, opened 2024-11-12
refactor: moving applications to a top level apps directory

sminez:
Other than the relative path imports in some of the test files this should be a straight move of the three application packages out to `/apps`. I've also added some initial notes on the proposed structure that I'm aiming to move things to in follow up PRs.


matthew-hagemann (approved):
LGTM!

d-loose (approved):

PR #869, opened 2024-11-19
chore(deps): track subiquity main

dbungert:
LP: #2088978 shows a SerializationError related to parsing of the new install-sources format, which includes information about what kernel to use.

Pick up subiquity main again, which we should do anyway but also because it knows how to work with the new install-sources format.


dbungert:
I have done a test build injecting subiquity main into the plucky daily desktop ISO and it works fine.  

I don't have the ability to set the snap label, so I cannot make the "Check snap label" test pass.

seb128 (approved):
Thanks

PR #870, opened 2024-11-20
ci: add write permissions for the snap release job

d-loose:
For pull requests raised from forked repositories, the `GITHUB_TOKEN` only comes with `read` permissions by default. This PR adds `write` permissions for the repository contents, so that the job can create/modify the release branch.


sminez (approved):
As discussed on Mattermost, this should be fine given that the job runs following PR merges rather than directly on PRs raised from arbitrary forks :+1:

PR #876, opened 2024-11-21
feat!: update Flutter and dependencies

d-loose:
Updates Flutter to 3.24.5 and Yaru to ^5.3.0.

P.S. after sorting out all the linter warnings concerning deprecated methods, there don't seem to be any visual changes after all.

UDENG-5159

### DIFF
--- a/packages/ubuntu_provision/lib/src/eula/eula_page.dart
+++ b/packages/ubuntu_provision/lib/src/eula/eula_page.dart
@@ -23,23 +23,19 @@ final eulaPageProvider = FutureProvider<File>((ref) async {
   final localizedEula = File(p.join(directory, 'EULA_$locale.pdf'));
   late final fallbackEula = File(p.join(directory, 'EULA.pdf'));
   final eulaFile = localizedEula.existsSync() ? localizedEula : fallbackEula;
+  assert(eulaFile.existsSync()); // throw error if fallback file doesn't exist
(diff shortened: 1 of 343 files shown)


packages/ubuntu_provision/lib/src/eula/eula_page.dart
```diff
@@ -23,39 +23,27 @@ final eulaPageProvider = FutureProvider<File>((ref) async {
   final localizedEula = File(p.join(directory, 'EULA_$locale.pdf'));
   late final fallbackEula = File(p.join(directory, 'EULA.pdf'));
   final eulaFile = localizedEula.existsSync() ? localizedEula : fallbackEula;
+  assert(eulaFile.existsSync()); // throw error if fallback file doesn't exist
```
sminez, line 26: is this refactor something that is required as part of updating the Flutter version? Not saying we shouldn't do it but it feels like this its own thing.

d-loose: I didn't have the time to debug it thoroughly, but the integration test failed since the `eulaPageProvider` was stuck in a loading state when trying to read a non-existent file. Making the page itself a stateless widget and adding this assertion fixed the behavior, but I didn't find the underlying cause for the test failure (instead of going down a hole of subtle UI state issues, I decided to simplify it a bit)

sminez (approved):

PR #882, opened 2024-11-29
feat: landscape magic attach

matthew-hagemann:
WIP: Currently just lifting the Ubuntu Pro pages from init out into bootstrap.


matthew-hagemann:
@d-loose still a WIP, and haven't removed the extra pages yet,  but if you are happy with how the landscape page and model are looking and how I unblocked the tests, I'll close the ticket to move over to riverpod annotations. Thanks again for all the help!

d-loose:
One thing I forgot to mention: we briefly talked about https://pub.dev/packages/qr_flutter the other day - would be great if you could check whether that's a more future-proof package for QR code generation than the one we're currently using here

matthew-hagemann:
> One thing I forgot to mention: we briefly talked about https://pub.dev/packages/qr_flutter the other day - would be great if you could check whether that's a more future-proof package for QR code generation than the one we're currently using here

Discussed on MM, we found the current library produced (bottom) clearer QR codes, so will not swap libraries.
image

apps/ubuntu_bootstrap/lib/pages/landscape/landscape_page.dart
```diff
@@ -0,0 +1,98 @@
...
+        leading: const BackWizardButton(),
+        // Trailing to be removed in the future to match designs. Leaving this here for now for easier debugging.
+        trailing: [
+          WizardButton(
+            label: UbuntuLocalizations.of(context).nextLabel,
+            onActivated: Wizard.of(context).next,
+          ),
+        ],
+      ),
+      imageTitleWidget: Column(
```
matthew-hagemann, line 37: I see this is the widget we use for positioning of all our illustrations. I put the QR code here to match the designs and the Horizontal page easily renders it where I want it, but not 100% sure its the appropriate place given the title is `imageTitleWidget`?

apps/ubuntu_bootstrap/lib/l10n/ubuntu_bootstrap_en.arb
```diff
@@ -925,5 +925,17 @@
         "type": "String"
       }
     }
-  }
+  },
+  "validate": "Validate",
```
d-loose, line 929: accidentally re-added during rebase
```suggestion
```

apps/ubuntu_bootstrap/test/installer_wizard_test.dart
```diff
@@ -64,6 +64,8 @@ void main() {
   tearDown(resetAllServices);
 
   testWidgets('try ubuntu', (tester) async {
+    // final landscapeModel = buildLandscapeModel();
+    // final tokenNotifier = buildTokenNotifier();
```
d-loose, line 68: leftover comments

apps/ubuntu_bootstrap/test/test_utils.dart
```diff
@@ -96,6 +96,10 @@ MockPageConfigService setupMockPageConfig({
               (step) => MapEntry(step.name, const PageConfigEntry()),
             ),
       );
+
+  // Debug log the pages being configured
+  debugPrint('Configured Pages: ${pages.keys.join(', ')}');
```
d-loose, line 101: debug statement

d-loose:
Looks really good so far!
One thing I noticed during a quick test with your mock server: if I change my mind and navigate back from the landscape page that shows the QR after entering an fqdn, the listener on the stream returned by `watch()` is never cancelled, so if a successful response comes back after some time, we still write the autoinstall file and restart the installer. I think that's not supposed to happen, so we should make sure we only listen to the response stream while the user is on that page.

More detailed review comments soon!

apps/ubuntu_bootstrap/lib/app/installation_step.dart
```diff
@@ -41,6 +43,8 @@ enum InstallationStep with RouteName {
   refresh(RefreshPage.new, allowedToHide: true),
   tryOrInstall(TryOrInstallPage.new, discreteStep: false, allowedToHide: true),
   autoinstall(AutoinstallPage.new, allowedToHide: true),
+  autoinstallLandscapeDomain(LandscapeDomainPage.new),
+  landscape(LandscapePage.new, required: true),
```
d-loose, line 47: Can we make all the enums and page names consistent (`AutoinstallLandscape...`)?

apps/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_landscape_domain_page.dart
```diff
@@ -0,0 +1,100 @@
...
+
+class _NextButton extends ConsumerWidget {
+  const _NextButton();
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final landscapeModel = ref.watch(landscapeDataModelProvider);
+    final theme = Theme.of(context);
+    final l10n = UbuntuBootstrapLocalizations.of(context);
+//    final l10n = UbuntuBootstrapLocalizations.of(context);
```
d-loose, line 67:
```suggestion
```

apps/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_landscape_model.dart
```diff
@@ -0,0 +1,124 @@
...
+import 'package:ubuntu_bootstrap/app/installer_model.dart';
+import 'package:ubuntu_bootstrap/pages/loading/loading_provider.dart';
+import 'package:ubuntu_bootstrap/services.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:yaml/yaml.dart';
+
+part 'autoinstall_landscape_model.freezed.dart';
+part 'autoinstall_landscape_model.g.dart';
+
+final _log = Logger();
```
d-loose, line 15: The logger should be named

d-loose (commented):
I have a bit of a refactoring proposal (since you've already spent quite a lot of time on this PR, I wouldn't mind merging this and coming back to it later on, though).

Current state management:
There's a single view model (`LandscapeDataModel`) for both the 'landscape domain' page where `attach()` is called and the 'landscape' page where `watch()` is called. The state this model holds includes the domain entered by the user as well as the status response from the server. It also keeps track of a stream/provider subscription and updates its own state whenever a stream event is received (after calling `watch()`). Listening/cancelling needs to be handled explicitly through corresponding method calls.


Proposed refactor:
We can split this into two view models. The one for the 'landscape domain' page would
- hold the domain entered by the user
- provide an `attach()` method

Instead of storing the received `userCode` in the view model, we could simply pass it to the route that displays the 'landscape' page.

The model for the 'landscape' page would then be a stream provider family (i.e. a stream provider that receives the `userCode` as an argument to its `build` function) that exposes a stream of UI states for the page. Simply using `ref.watch(autoinstallLandscapeProvider(userCode))` would then take care of the stream's lifecycle (i.e. start listening when the user enters the page, cancel when they leave).

-> UDENG-5870

apps/ubuntu_bootstrap/integration_test/ubuntu_bootstrap_test.dart
d-loose, line 1: Could we remove the added empty lines in this file?

apps/ubuntu_bootstrap/lib/app/installation_step.dart
```diff
@@ -41,6 +43,8 @@ enum InstallationStep with RouteName {
   refresh(RefreshPage.new, allowedToHide: true),
   tryOrInstall(TryOrInstallPage.new, discreteStep: false, allowedToHide: true),
   autoinstall(AutoinstallPage.new, allowedToHide: true),
+  autoinstallLandscapeDomain(LandscapeDomainPage.new),
+  autoinstallLandscapeQR(LandscapePage.new, required: true),
```
d-loose, line 47: Can we make the naming consistent? (the file names should be adjusted correspondigly)
```suggestion
  autoinstallLandscapeDomain(AutoinstallLandscapeDomainPage.new),
  autoinstallLandscapeQr(AutoinstallLandscapeQRPage.new, required: true),
```

apps/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_landscape_page.dart
```diff
@@ -0,0 +1,149 @@
...
+                      ),
+                      const SizedBox(width: kWizardSpacing / 2),
+                      switch (model.authenticationStatus) {
+                        service.AuthenticationStatus.errorCodeExpired => Row(
+                            mainAxisSize: MainAxisSize.min,
+                            children: [
+                              const ErrorIcon(),
+                              const SizedBox(width: kWizardSpacing / 4),
+                              Text(
+                                'Code expired, please try again',
```
d-loose, line 94: This string should be translatable

apps/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_landscape_page.dart
```diff
@@ -0,0 +1,149 @@
...
+                        service.AuthenticationStatus.authenticationSuccess ||
+                        service.AuthenticationStatus.authenticationPending =>
+                          SizedBox.shrink(),
+                        _ => Row(
+                            mainAxisSize: MainAxisSize.min,
+                            children: [
+                              const ErrorIcon(),
+                              const SizedBox(width: kWizardSpacing / 4),
+                              Text(
+                                'Login failed, please try again',
```
d-loose, line 114: Same here, this string should be translatable

apps/ubuntu_bootstrap/test/test_utils.dart
```diff
@@ -96,6 +96,7 @@ MockPageConfigService setupMockPageConfig({
               (step) => MapEntry(step.name, const PageConfigEntry()),
             ),
       );
+
```
d-loose, line 99:
```suggestion
```

apps/ubuntu_bootstrap/lib/pages/autoinstall/autoinstall_landscape_page.dart
matthew-hagemann: Ah sorry, missed these!

apps/ubuntu_bootstrap/lib/app/installation_step.dart
```diff
@@ -41,6 +41,8 @@ enum InstallationStep with RouteName {
   refresh(RefreshPage.new, allowedToHide: true),
   tryOrInstall(TryOrInstallPage.new, discreteStep: false, allowedToHide: true),
   autoinstall(AutoinstallPage.new, allowedToHide: true),
+  autoinstallLandscapeDomain(AutoinstallLandscapeDomainPage.new),
+  autoinstallLandscapeQr(AutoinstallLandscapeQrPage.new, required: true),
```
d-loose, line 45: Sorry, missed this earlier: is there a reason this is marked as `required`? I think we should add it to the default `whitelabel.yaml` instead (the `image` can be left blank).

matthew-hagemann: Ah no, I think that was from testing originally, will remove the required :+1: 

When you say add it to the `whitelabel.yaml`, do you mean the one in `packages/ubuntu_provision/assets/whitelabel.yml`?

d-loose (approved):

PR #884, opened 2024-12-02
docs: adding dev-docs for adding a new page to the installer

matthew-hagemann:
Based on my learnings from stubbing out landscape integratino in #882, I have put together some `dev-docs` outlining the broad steps required for adding a new page to the insatller, as well as adding some doc comments on some relevant dart components that did not have any.

### DIFF
--- a/apps/ubuntu_bootstrap/lib/app/installation_step.dart
+++ b/apps/ubuntu_bootstrap/lib/app/installation_step.dart
@@ -15,9 +15,9 @@ import 'package:ubuntu_wizard/ubuntu_wizard.dart';
 /// New pages must be added as additional enums to integrate into the workflow.
 ///
 /// Properties:
-/// - `discreteStep`: Indicates whether the step is displayed as a discrete wizard step. Discrete 
+/// - `discreteStep`: Indicates whether the step is displayed as a discrete wizard step. Discrete
 /// steps get their own circle progress indicator at the bottom of the window.
-/// - `wizardStep`: Specifies if the step is part of the wizard flow. Wizard steps appear with 
+/// - `wizardStep`: Specifies if the step is part of the wizard flow. Wizard steps appear with
 /// circle progress indicators and the default step navigation at the bottom of the window.
 /// - `required`: Marks the step as mandatory in the workflow.
 /// - `allowedToHide`: Determines if the step can be conditionally hidden.
--- a/apps/ubuntu_bootstrap/lib/app/installation_step.dart
+++ b/apps/ubuntu_bootstrap/lib/app/installation_step.dart
@@ -15,10 +15,17 @@ import 'package:ubuntu_wizard/ubuntu_wizard.dart';
 /// New pages must be added as additional enums to integrate into the workflow.
 ///
 /// Properties:
-/// - `discreteStep`: Indicates whether the step is displayed as a discrete wizard step.
-/// - `wizardStep`: Specifies if the step is part of the wizard flow.
+/// - `discreteStep`: Indicates whether the step is displayed as a discrete wizard step. Discrete 
+/// steps get their own circle progress indicator at the bottom of the window.
+/// - `wizardStep`: Specifies if the step is part of the wizard flow. Wizard steps appear with 
+/// circle progress indicators and the default step navigation at the bottom of the window.
 /// - `required`: Marks the step as mandatory in the workflow.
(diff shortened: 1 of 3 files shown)


apps/ubuntu_bootstrap/lib/app/installation_step.dart
```diff
@@ -6,6 +6,19 @@ import 'package:ubuntu_provision/ubuntu_provision.dart';
...
+/// Represents each step or page in the installer workflow.
+///
+/// Each step corresponds to an enum value, which defines the page's behavior
+/// and its role in the installer. The order of the enums determines the sequence
+/// in which the steps appear in the installer.
+///
+/// New pages must be added as additional enums to integrate into the workflow.
+///
+/// Properties:
+/// - `discreteStep`: Indicates whether the step is displayed as a discrete wizard step.
```
sminez, line 18: we probably want to explain what being a discrete wizard step means here.

dev-docs/adding-a-new-page-to-the-installer.md
```diff
@@ -0,0 +1,55 @@
+# Adding a new page to the installer
+
+## Add your page
+- Under `apps/ubuntu_bootstrap/lib/pages`, create a new directory for the page you would like to add
+ie: `apps/ubuntu_bootstrap/lib/pages/<your-feature>`
+- Create at minimum the following files:
+	- `<your-feature>_page.dart`: Defines the UI for the page you would like to add to the installer.
+	- `<your-feature>_model.dart`: Implements the provider and integrates with backend services to
```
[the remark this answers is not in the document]
sminez: that'll probably get out of date over time, better to describe the functionality split between the files and direct people to refer to existing pages to see what that looks like

dev-docs/adding-a-new-page-to-the-installer.md
```diff
@@ -0,0 +1,55 @@
...
+## Add your page
+- Under `apps/ubuntu_bootstrap/lib/pages`, create a new directory for the page you would like to add
+ie: `apps/ubuntu_bootstrap/lib/pages/<your-feature>`
+- Create at minimum the following files:
+	- `<your-feature>_page.dart`: Defines the UI for the page you would like to add to the installer.
+	- `<your-feature>_model.dart`: Implements the provider and integrates with backend services to 
+    support your page's functionality.
+- Export your page for use in the installer by adding it to `apps/ubuntu_bootstrap/lib/pages.dart`
+- Add your page to the installation steps by adding it as an enum to `InstallationStep` in 
+`apps/ubuntu_bootstrap/lib/app/installation_step.dart`. The order of the `InstallationStep` enums
```
sminez, line 12: do we need to call out the weird behaviour of how page transitions can affect state here? @d-loose I still don't think I'm 100% sure I have my head round what the current behaviour is other than "here be dragons"

d-loose: Not exactly sure what you're referring to - do you mean the fact that pages come with a `load()` method that determines whether it's skipped or not (and additionally there's some logic in the `PageConfigService` that determines whether or not a page is shown based on the whitelabel.yaml)?

dev-docs/adding-a-new-page-to-the-installer.md
```diff
@@ -0,0 +1,55 @@
+# Adding a new page to the installer
+
+## Add your page
+- Under `apps/ubuntu_bootstrap/lib/pages`, create a new directory for the page you would like to add
+ie: `apps/ubuntu_bootstrap/lib/pages/<your-feature>`
```
d-loose, line 5: Depending on how general the page is and whether we want to use it in `ubuntu_init` or elsewhere later on, the page could be added to `ubuntu_provision` instead

dev-docs/adding-a-new-page-to-the-installer.md
sminez: Maybe I'm mis-remembering but I thought there was cakllbacks that fired to update state as you moved between pages and that in some places pages know (need to know?) what page is navigating to them.

matthew-hagemann: My understanding is the gotcha here is when you have sections with their own wizard that share a single model and state is persisted within that model between wizard steps? 

ie: `pages/storage/storage_wizard.dart`

d-loose (approved):

PR #892, opened 2024-12-11
feat: Erase and install

sminez:
This is work in progress for adding support for "Erase & Install" based the Subiquity work being done [here](https://github.com/canonical/subiquity/pull/2117). We wont be able to land this before [this bug](https://bugs.launchpad.net/subiquity/+bug/2091172) around logical partitions being re-labelled and breaking the mapping between installed OS and partition is fixed.

There are a still few UI quirks that need sorting out before this is production ready but the core functionality is now in place:

- [x] Selecting a guided target that requires scrolling on the "Disk Setup" page results in the list of targets jumping to the top, hiding the selection.
- [x] Hovering over a target that is partially scrolled off the bottom of the view results in broken background colors as the partially obscured widget applies its background color incorrectly (see second screenshot below).
- [ ] Text for the "Disk setup" section on the "Ready to Install" page can end up wrapping in an ugly way for erase & install targets (see third screenshot below).

cc. @anasereijo

guided-targets

broken-background

confirmation

UDENG-5549

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,6 @@
 [submodule "subiquity"]
 	path = packages/subiquity_client/subiquity
-	url = https://github.com/canonical/subiquity.git
-	branch = main
+	; url = https://github.com/canonical/subiquity.git
+	; branch = main
+	url = https://github.com/ogayot/subiquity.git
+	branch = erase-and-install
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,4 @@
 [submodule "subiquity"]
 	path = packages/subiquity_client/subiquity
-	; url = https://github.com/canonical/subiquity.git
-	; branch = main
-	url = https://github.com/ogayot/subiquity.git
-	branch = erase-and-install
+	url = https://github.com/canonical/subiquity.git
+	branch = main
--- a/apps/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
+++ b/apps/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
@@ -8,10 +8,17 @@ import 'package:ubuntu_bootstrap/pages/storage/storage_page.dart';
 import 'package:ubuntu_bootstrap/services.dart';
 import 'package:ubuntu_provision/ubuntu_provision.dart';
 
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+
+final _log = Logger('storage-model');
+
 /// Available storage types.
 enum StorageType {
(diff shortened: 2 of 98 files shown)


.gitmodules
```diff
@@ -1,4 +1,6 @@
 [submodule "subiquity"]
 	path = packages/subiquity_client/subiquity
-	url = https://github.com/canonical/subiquity.git
-	branch = main
+	; url = https://github.com/canonical/subiquity.git
+	; branch = main
+	url = https://github.com/ogayot/subiquity.git
+	branch = erase-and-install
```
sminez, line 6: Required for initial testing against the in-progress work on the Subiquity side. This needs to be removed before merging.

anasereijo:
@sminez some small things:

- [ ] typo in "permanently"
- [ ] can we make the word "warning" orange or whatever the warning colour is in yaru?
- [ ] first option could read: "Install alongside existing operating systems" (remind me, it is what this does, correct?)
- [ ] can we say "install Ubuntu [version]" instead of just saying Ubuntu?

sminez:
@anasereijo this is the current UI at this point.

apps/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
```diff
@@ -187,14 +244,15 @@ class StorageModel extends SafeChangeNotifier {
       return 'resize_use_free';
     } else if (type == StorageType.manual) {
       return 'manual';
+    } else if (type case StorageTypeEraseInstall()) {
+      return 'reuse_partition';
```
d-loose, line 248: `reuse_partition` is probably not the right label here. The TODO note refers to an installation scenario we don't support (re-using an existing ubuntu installation and keeping the user's `/home` directory).
The label is only being used for the telemetry that is written to `/var/log/installer/telemetry`

apps/ubuntu_bootstrap/lib/pages/storage/storage_page.dart
```diff
@@ -45,62 +51,125 @@ class StoragePage extends ConsumerWidget with ProvisioningPage {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final model = ref.watch(storageModelProvider);
+    // final model = ref.watch(storageModelProvider);
```
d-loose, line 54:
```suggestion
```

anasereijo:
🚀 🚀 MERGEEEE 🚀 🚀

d-loose (approved):

bittin:
Swedish updated :)

PR #910, opened 2025-01-10
feat: implement new autoinstall UI

d-loose:
Breaks up the existing autoinstall page into multiple steps and adds a file picker for local paths.

Screenshot From 2025-01-10 16-53-59
Screenshot From 2025-01-10 16-54-09
Screenshot From 2025-01-10 16-54-26
Screenshot From 2025-01-10 16-56-09
Screenshot From 2025-01-10 16-58-10

- Extracts non-UI logic into `AutoinstallService`
- `AutoinstallModel` now only holds the state of the autoinstall type: `none` (interactive), `direct` (load file from URL or local path), `landscape` (obtain file via landscape). It also provides a method to restart the UI.
- New `AutoinstallDirectModel` holds the UI state of the URL text box and file picker which have been moved to a separate page. It also provides a method to show the file picker (via xdg desktop portals)

Error handling still needs improving, but I'd do that in a follow-up PR.

UDENG-5748


apps/ubuntu_bootstrap/lib/services/autoinstall_service.dart
```diff
@@ -0,0 +1,98 @@
...
+  final SubiquityClient _subiquity;
+  final SubiquityServer _subiquityServer;
+  final HttpClient _httpClient;
+  final FileSystem _fs;
+  final bool _liveRun;
+
+  static const filename = 'autoinstall.yaml';
+  static const targetDir = '/';
+
+  Future<void> writeFile(String content) async {
```
matthew-hagemann, line 34: When importing an `autoinstall.yaml` from a file, the file is now always present. If I exit and restart the installer, I'm locked into that `autoinstall.yaml` with now way of changing it without manually deleting the file. Not sure if we want a way to clear it in the `Ready to install` screen?

d-loose: As discussed during standup, this is currently out of scope, but a very sensible feature request

matthew-hagemann (approved):

bittin:
Now in Swedish

PR #915, opened 2025-01-23
feat: supporting setting a passphrase with TPM FDE

sminez:
Allow the user to specify a passphrase for TPM FDE (but do not _require_ that they set one) and overhaul the recovery key information page to match the new Figma designs.

This is still a work in progress for now while we wait for the Subiquity API work to be completed. One outstanding piece of functionality that needs to be implemented is making use of the `/storage/v2/calculate_entropy` endpoint when it is available to provide feedback on the user's passphrase strength.

UDENG-5816

### DIFF
--- a/apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
+++ b/apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
@@ -50,7 +50,7 @@ class PassphraseModel extends SafeChangeNotifier {
       passphrase.isNotEmpty && passphrase == confirmedPassphrase;
 
   /// Whether the user can skip providing a passphrase
-  bool get canSkip => isTpm ? passphrase.isEmpty : false;
+  bool get canSkip => isTpm && passphrase.isEmpty;
 
   /// Initializes the model.
   Future<bool> init() async {
--- a/apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
+++ b/apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
@@ -18,7 +18,6 @@ class PassphraseModel extends SafeChangeNotifier {
       _passphrase,
       _confirmedPassphrase,
       _showPassphrase,
-      _isTpm,
     ]).addListener(notifyListeners);
   }
 
@@ -27,7 +26,6 @@ class PassphraseModel extends SafeChangeNotifier {
   final _passphrase = ValueNotifier('');
   final _confirmedPassphrase = ValueNotifier('');
   final _showPassphrase = ValueNotifier(false);
-  final _isTpm = ValueNotifier(false);
 
   /// The current passphrase.
   String get passphrase => _passphrase.value;
@@ -42,8 +40,10 @@ class PassphraseModel extends SafeChangeNotifier {
   set showPassphrase(bool value) => _showPassphrase.value = value;
 
   /// Defines whether or not the user has selected TPM FDE
-  bool get isTpm => _isTpm.value;
-  set isTpm(bool value) => _isTpm.value = value;
+  bool get isTpm => [
+        GuidedCapability.CORE_BOOT_ENCRYPTED,
+        GuidedCapability.CORE_BOOT_PREFER_ENCRYPTED,
+      ].contains(_service.guidedCapability);
 
   /// Whether the current input is valid.
   bool get isValid =>
@@ -57,18 +57,14 @@ class PassphraseModel extends SafeChangeNotifier {
     switch (_service.guidedCapability) {
       case GuidedCapability.LVM_LUKS:
       case GuidedCapability.ZFS_LUKS_KEYSTORE:
-        isTpm = false;
-
       case GuidedCapability.CORE_BOOT_ENCRYPTED:
       case GuidedCapability.CORE_BOOT_PREFER_ENCRYPTED:
-        isTpm = true;
+        await loadPassphrase();
+        return true;
 
       default:
         return false;
     }
-
-    await loadPassphrase();
-    return true;
   }
 
   /// Loads the passphrase from the service.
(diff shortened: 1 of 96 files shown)


apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
d-loose, line 1: I think we could simplify this a bit, since we don't need `isTpm` to be externally mutable. It could just be a getter that tells us whether the selected guided capability is `CORE_BOOT_*_ENCRYPTED`:

```dart
    bool get isTpm => [
        GuidedCapability.CORE_BOOT_ENCRYPTED,
        GuidedCapability.CORE_BOOT_PREFER_ENCRYPTED,
      ].contains(_service.guidedCapability);
```

apps/ubuntu_bootstrap/lib/pages/storage/storage_routes.dart
```diff
@@ -6,5 +6,6 @@ enum StorageSteps with RouteName {
   guidedResize,
   manual,
   passphrase,
+  tpmPassphrase,
```
d-loose, line 9: Should be removed now that we have a generic passphrase page
```suggestion
```

apps/ubuntu_bootstrap/test/storage/recovery_key/test_recovery_key.dart
```diff
@@ -10,5 +10,6 @@ export 'test_recovery_key.mocks.dart';
 RecoveryKeyModel buildRecoveryKeyModel({bool? init}) {
   final model = MockRecoveryKeyModel();
   when(model.init()).thenAnswer((_) async => init ?? false);
+  when(model.confirmed).thenReturn(true);
```
d-loose, line 13: Not sure if it's worth having a minimal test that checks whether or not the 'next' button is enabled depending on the checkobox state.

sminez: Maybe, but the logic is literally just a direct check of the boolean value so all it would be testing is that the checkbox's value is `model.confirmed` :shrug: 
```dart
          NextWizardButton(
            enabled: ref.watch(
              recoveryKeyModelProvider.select((model) => model.confirmed),
            ),
          ),
```

d-loose (dismissed):

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
```diff
@@ -39,18 +39,32 @@ class PassphraseModel extends SafeChangeNotifier {
...
+        GuidedCapability.CORE_BOOT_ENCRYPTED,
+        GuidedCapability.CORE_BOOT_PREFER_ENCRYPTED,
+      ].contains(_service.guidedCapability);
+
   /// Whether the current input is valid.
   bool get isValid =>
       passphrase.isNotEmpty && passphrase == confirmedPassphrase;
 
+  /// Whether the user can skip providing a passphrase
+  bool get canSkip => isTpm ? passphrase.isEmpty : false;
```
d-loose, line 53: Nitpick: I think `isTpm && passphrase.isEmpty` is easier to understand

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_page.dart
```diff
@@ -21,29 +21,48 @@ class PassphrasePage extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final isTpm = ref.watch(
+      passphraseModelProvider.select((model) => model.isTpm),
+    );
     final lang = UbuntuBootstrapLocalizations.of(context);
+    final l10n = UbuntuLocalizations.of(context);
```
d-loose, line 28: Could we call this `ulang` for consistency with other pages?

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_widgets.dart
```diff
@@ -6,6 +6,8 @@ import 'package:ubuntu_bootstrap/pages/storage/passphrase/passphrase_model.dart'
 import 'package:ubuntu_widgets/ubuntu_widgets.dart';
 import 'package:yaru/icons.dart';
 
+// TODO: this will need to be conditionally hooked up to the /storage/v2/calculate_entropy
+//       endpoint in Subiquity once it is available
```
d-loose, line 10: I guess that also means we'll do the `subiquity_client` changes to support the new endpoint in a separate PR?

sminez: Yes, I shared a quick update in the team channel in Mattermost: the snapd API isn't ready to integrate with yet so they would prefer we don't hook anything up at all (even the stub being provided by Subiquity).

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_model.dart
sminez: Yeah that looks good.

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_page.dart
sminez: I've not seen `ulang` in other pages in the installer but I can change it if that's what you'd prefer?

d-loose: Right, seems like the installer localizations are rarely used in the same place as those coming from `ubuntu_localizations`, what I had in mind is actually part of the test utils: https://github.com/canonical/ubuntu-desktop-provision/blob/1cd4ada4883d3773e8bd0950f0f25ad2c28e1d19/packages/ubuntu_provision/test/test_utils.dart#L22-L30

d-loose (approved):

PR #917, opened 2025-01-28
Advanced storage options

matthew-hagemann:
Breaking advanced storage options into its own page. Continuation of #858 

Screenshot From 2025-02-06 16-18-11
Screenshot From 2025-02-07 13-53-48
Screenshot From 2025-02-07 13-53-25


matthew-hagemann:
@anasereijo Let me know your thoughts on how the designs came out :slightly_smiling_face: I wasn't sure re: the `hide` button to collapse the advanced options

matthew-hagemann:
Will open a PR to patch Go separately and link it back here when its done:

matthew-hagemann:
Screen recording of the interaction :slightly_smiling_face:  @anasereijo

[Screencast From 2025-02-07 14-10-53.webm](https://github.com/user-attachments/assets/2e593d25-7056-495d-9c2c-783167a01666)

anasereijo:
@matthew-hagemann I've been thinking about the use of the 'experimental' chip, and there's no way to avoid wrapping it (especially with translations). that said, I think we should wrap it properly to the next line aligned to the left

- the first option 'no encryption' doesn't have support text anymore

everything else is looking good! the interactions are working really nicely 👍

anasereijo:
@matthew-hagemann actually, I have been thinking about this, and I wonder if it's possible to do something more like this instead:

Screenshot 2025-02-10 at 11 10 16

so the chip/label will be kept always at the top right corner. When the title is too long, wrap it to a second line.

apps/ubuntu_bootstrap/lib/pages/storage/storage_guided_capabilities_page.dart
```diff
@@ -0,0 +1,196 @@
...
+    final lang = UbuntuBootstrapLocalizations.of(context);
+    final flavor = ref.watch(flavorProvider);
+    final experimentalBadgeText =
+        UbuntuBootstrapLocalizations.of(context).installationTypeExperimental;
+
+    final target = model
+        .getAllTargets()
+        .firstWhereOrNull((t) => t is GuidedStorageTargetReformat);
+
+    if (target == null) return const SizedBox();
```
d-loose, line 155: We will need to generalize this logic here, since we can now see this option while a different guided target (one that is not `GuidedStorageTargetReformat`) is selected.

I've tested this with the `win10-along-ubuntu` machine config, where we get a `Reformat` and two `Resize` targets. The latter don't include the `CORE_BOOT_PREFER_ENCRYPTED` capability in neither `allowed` nor `disallowed`, so I think we should hide the option entirely in this case (currently it's shown and enabled, so continuing with the installation will lead to a crash, since we're sending a guided target response that includes an unsupported capability). At the same time  we should make sure that the option is always present if it's included in one of those, but keep it disabled if the capability is disallowed.

apps/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
```diff
@@ -71,12 +72,18 @@ class StorageModel extends SafeChangeNotifier {
...
     this._product,
     this._session,
-  );
+  ) : showAdvanced = false;
 
   final StorageService _storage;
   final TelemetryService? _telemetry;
   final ProductService _product;
   final SessionService _session;
+  bool showAdvanced;
```
d-loose, line 81: Nitpick: could we make that private to prevent accidentally changing it without calling `notifyListeners()`?
```suggestion
  var _showAdvanced = false;
  
  bool get showAdvanced => _showAdvanced;
```

d-loose: On second though: maybe this doesn't need to be part of the storage model at all? This is only needed for the UI state of the new page, right?

apps/ubuntu_bootstrap/lib/pages/storage/storage_page.dart
```diff
@@ -256,7 +207,7 @@ class _InstallationTypeTile extends ConsumerWidget {
           title: title,
           subtitle: Column(
             crossAxisAlignment: CrossAxisAlignment.start,
-            children: [subtitle, trailing].whereNotNull().toList(),
+            children: [subtitle].whereNotNull().toList(),
```
d-loose, line 210: We can get rid of the entire column here and simply use `subtitle: subtitle`.

apps/ubuntu_bootstrap/lib/pages/storage/storage_guided_capabilities_page.dart
d-loose, line 1: Can we rename this file to `guided_capabilities_page.dart` for consistency?

apps/ubuntu_bootstrap/test/storage/storage_page_test.dart
```diff
@@ -32,6 +33,20 @@ void main() {
     );
   }
 
+  Widget buildAdvancedPage(StorageModel model) {
```
d-loose, line 36:
```suggestion
  Widget buildGuidedCapabilitiesPage(StorageModel model) {
```

apps/ubuntu_bootstrap/lib/pages/storage/storage_model.dart
sminez: yes actually, its just tracking whether or not the user has clicked to expand out to show the full list of advanced options.

d-loose (approved):
@sminez thanks for the fixes - everything looks good to me now!
I've added a very simple integration test that tests LVM+LUKS in a dual boot scenario

matthew-hagemann:
@d-loose @sminez thank you for all the help getting this done! :rocket:

sminez (approved):

bittin:
Swedish updated

anasereijo:
@matthew-hagemann did you manage to check my suggestion for the labels?

juanruitina:
We implemented this in the end as suggested by @anasereijo: https://github.com/canonical/ubuntu-desktop-provision/pull/917#issuecomment-2647670172

PR #938, opened 2025-02-06
chore: setting up tics ci

sminez:
Being tracked internally under UDENG-5819

Currently we are unable to get the TiCS integration working as there is broken config on their side that we need to chase them up about.


d-loose:
Closing this in favor of #1042

PR #943, opened 2025-02-11
feat: don't allow modification of unsupported disks

d-loose:
UI changes to support https://github.com/canonical/subiquity/pull/2157

Screencast using probe data and curtin as mentioned in the PR above:

[Screencast From 2025-02-11 11-34-16.webm](https://github.com/user-attachments/assets/663e11f3-bc98-467c-b908-93147bb14fa3)

In this case `/dev/sda` is marked as `unsupported`, so the UI blocks adding/removing/modifying the disk's partitions and setting it as the boot device. Reformatting the entire disk works as usual, though.

@ogayot FYI

UDENG-5999

### DIFF
--- a/packages/subiquity_client/test/types_test.dart
+++ b/packages/subiquity_client/test/types_test.dart
@@ -276,7 +276,7 @@ void main() {
       password: '2',
       sizingPolicy: SizingPolicy.ALL,
       resetPartition: false,
-      pin: '0000',
+      pin: null,
       recoveryKey: null,
     );
     const json = <String, dynamic>{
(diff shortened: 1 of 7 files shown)


packages/subiquity_client/test/types_test.dart
```diff
@@ -268,17 +270,21 @@ void main() {
         diskId: '0',
         allowed: [],
         disallowed: [],
+        ptable: null,
       ),
       capability: GuidedCapability.LVM,
       password: '2',
       sizingPolicy: SizingPolicy.ALL,
       resetPartition: false,
+      pin: '0000',
```
sminez, line 279: why are we setting a PIN rather than a passphrase (given that we don't currently support PIN in the UI I mean)?

PR #964, opened 2025-02-28
chore(deps): stage python3-packaging for provision

dbungert:
We have added a dependency on python3-packaging for curtin, reflect that here.

Matching PR: https://github.com/canonical/subiquity/pull/2171


d-loose:
Thanks, I'm including this in #966

PR #970, opened 2025-03-10
fix: Resolve 1870620 - Mascot progress image contrast

ashuntu:
Adds a dim to the image when a download percentage is present. Also forces the download text to be white, since the image will always been dark anyway. New contrast ratio is around 5.

<details>
<summary>Screenshots</summary>

Without download percentage (before download starts, unchanged):
image

With download percentage (after download starts):
image
</details>

---

UDENG-5924

### DIFF
--- a/packages/ubuntu_provision/lib/src/widgets/mascot_avatar.dart
+++ b/packages/ubuntu_provision/lib/src/widgets/mascot_avatar.dart
@@ -26,13 +26,25 @@ class MascotAvatar extends StatelessWidget {
           ),
           child: image,
         ),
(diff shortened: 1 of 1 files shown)


d-loose:
Sorry, I didn't realize you were having a look at this as well - there was an internal discussion about this in [BRND-1769](https://warthogs.atlassian.net/browse/BRND-1769?focusedCommentId=695934) (currently still confidential as it includes the new mascot). Long story short - we decided to simply move the percentage below the mascot. I'll include this in the PR to update the mascot later today or tomorrow.

PR #974, opened 2025-03-12
Landscape UI tweaks

matthew-hagemann:
UI tweaks, including a dedicated error page to land on if the error returned from landscape is not retriable

Still need to do the refactor to make testing these pages, so there is not much to tweak in the tests currently


apps/ubuntu_bootstrap/lib/app/installation_step.dart
```diff
@@ -106,6 +110,9 @@ enum InstallationStep with RouteName {
         }
         return false;
       },
+      onBack: this == InstallationStep.autoinstallLandscapeError
+          ? (_) => InstallationStep.autoinstall.route
+          : null,
```
d-loose, line 115: Could this be a property on the `InstallationStep` enum that is `null` for every step but this one?

d-loose (approved):

bittin:
Swedish updated

PR #977, opened 2025-03-12
feat(resize-page): add disabled bitlocker dropdown entry

d-loose:
Adds disabled dropdown entries for bitlocker partitions on the 'guided resize' page.
See spec UD094.

Example with existing Ubuntu Plucky and a BitLocker partition, installing Ubuntu 24.10:
image

Again, since osprober doesn't give us proper information for the bitlocker partition, all we can do is display the partition type, instead of the name of the OS.


matthew-hagemann (dismissed):
Code LGTM!

juanruitina:
The disabled entry looks OK. One minor visual comment, likely off-scope for the issue: the "Restart into Windows" button in the info panel lacks a bit of spacing above and below, could you tweak? Something like whichever spacing the heading has would work.

d-loose:
Added some spacing:

image

matthew-hagemann (approved):

PR #979, opened 2025-03-12
fix: Wizard theme constrast adjustments

ashuntu:
This does a couple small things:

- Separates theme customizations into separate cursor, color, and shape functions
- Makes regular TextFields look like ValidatedTextFields (removed their background color)
- In high contrast mode, make subtitle text the same color as title text in list tiles

These are just the changes that could've been made in the theme itself (which isn't a lot), some of the other contrast changes will need some customization at a widget level.

<details>
<summary>Screenshots</summary>

**Removes background color on normal TextFields**
image

**High contrast mode list tiles**
image

</details>

---

UDENG-6291


juanruitina:
Fixes to keyboard testing field and subtitles in radio button panels look good to me, they seem to address issues (contrast between text and background is definitely >4.5:1)

d-loose (approved):
Thanks a lot!

PR #982, opened 2025-03-13
feat: Additional conditional messaging on landscape install, set min height on Yaru container

matthew-hagemann:


apps/ubuntu_bootstrap/lib/pages/confirm/confirm_page.dart
```diff
@@ -57,85 +57,107 @@ class ConfirmPage extends ConsumerWidget with ProvisioningPage {
...
+              lang.landscapeConfirmPageSuccessInfoContent,
+            ),
+          ),
+          const SizedBox(height: kWizardSpacing),
+        ],
+        Row(
+          children: [
+            Expanded(
+              child: YaruBorderContainer(
+                constraints: BoxConstraints(minHeight: 300),
```
matthew-hagemann, line 75: Not sure if there is a constant I should be using instead?

d-loose: I suppose if we wanted to always fill (at least) the entire height of the viewport, we'd have to handle the scrolling on this page manually (`HorizontalPage` has a `managedScrolling` property that can be set to `false`). We'd then need to wrap a `SingleChildScrollview` in a `LayoutBuilder` to obtain the available height and use an `IntrinsicHeight` within the scroll view - see "Expanding content to fit the viewport" in [these docs here](https://api.flutter.dev/flutter/widgets/SingleChildScrollView-class.html).
Not sure if you have time to go down that rabbit hole, though - I'd be fine with hardcoding an appropriate height for now.

d-loose (approved):

juanruitina:
Functionally, this looks OK. Only a minor visual comment: please reduce the line height of the notice's title. When wrapping, the two lines have too much space in between. We should fix this upstream at some point.

d-loose:
> Functionally, this looks OK. Only a minor visual comment: please reduce the line height of the notice's title. When wrapping, the two lines have too much space in between. We should fix this upstream at some point.

This is currently [hardcoded](https://github.com/ubuntu/yaru.dart/blob/0a1ee4f16a5efb332b9a62286f39d2ab6735cfb4/lib/src/widgets/yaru_info.dart#L172) in the `YaruInfoBox` widget. We could locally override it by providing a custom `titleTextStyle` where ever we use that widget if necessary.

matthew-hagemann:
> > Functionally, this looks OK. Only a minor visual comment: please reduce the line height of the notice's title. When wrapping, the two lines have too much space in between. We should fix this upstream at some point.
> 
> This is currently [hardcoded](https://github.com/ubuntu/yaru.dart/blob/0a1ee4f16a5efb332b9a62286f39d2ab6735cfb4/lib/src/widgets/yaru_info.dart#L172) in the `YaruInfoBox` widget. We could locally override it by providing a custom `titleTextStyle` where ever we use that widget if necessary.

Thanks! I've reduced it to 1.3 and updated the screenshot accordingly. My eyes think thats a good number?

d-loose (approved):

PR #993, opened 2025-03-20
fix: info badge alignment

d-loose:
Fixes the 'experimental' badge alignment issues as discussed with @juanruitina 

Before:

[Screencast From 2025-03-20 16-07-02.webm](https://github.com/user-attachments/assets/5e995490-05c5-4826-9d9c-fafe8dd5e290)

After:

[Screencast From 2025-03-20 16-08-11.webm](https://github.com/user-attachments/assets/432d703c-4b82-4954-9b78-246b83ada43b)


matthew-hagemann (approved):
Looks great! Thanks!

PR #997, opened 2025-03-21
chore: update go and dependencies

d-loose:
This should fix the dependabot alert https://github.com/canonical/ubuntu-desktop-provision/security/dependabot/7. I don't completely understand what dependabot is doing in #980. Since we're already using renovate to update dependencies and osv-scanner (#996) for vulnerability scanning, maybe we should simply get rid of dependabot?


matthew-hagemann:
`golangci-lint` seems to have very odd failure modes, but from my testing your issues and mine (kept getting typecheck errors on valid types) seem to happen when the version of `golangci-lint` is not targeting the version of `go` in use. I've bumped everything to `go` 1.24 and `golangci-lint` 1.64.8. 

Its running now, but finding new issues. The grpc work seems simple enough to swap away from the deprecated versions, so I think it would be better to get everything working against 1.24. I've run out of time now but will look at fixing those tomorrow morning :slightly_smiling_face:

provd/.golangci.yaml
```diff
@@ -7,7 +7,6 @@ linters:
     - durationcheck
     - errname
     - errorlint
-    - exportloopref
```
matthew-hagemann, line 10: Now depricated

provd/internal/daemon/daemon_test.go
```diff
@@ -303,6 +303,7 @@ func createClientConnection(t *testing.T, socketPath string) (success bool, disc
 	go func() {
 		defer close(connected)
 		var err error
+		//nolint:staticcheck //This is still supported through 1.x., and we may depricated this before then.
```
matthew-hagemann, line 306: This is the only place where I've ignored the lint. There is larger effort to fix this as as `grpc.WithBlock()` is also deprecated. If we end up still wanting to move forward with `provd` running as a daemon, we can apply the [fixes](https://github.com/ubuntu/authd/commit/14e547bf1ea091699e65be70504e3c9718dcf257) that they used in `authd`.

d-loose (commented):
Thanks for the help - you changes look good to me!

Issue #1011, opened 2025-03-31
Confusing line wrapping on 'Review your choices' screen


vanvugt:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Bug Description

There is confusing line wrapping on the 'Review your choices' screen. What should be one line per list item is occupying two lines.

![Image](https://github.com/user-attachments/assets/ea7e1696-82d9-4ad0-a4aa-62acf435f1e1)

### Steps to reproduce the behavior

Install Plucky on a high DPI (say >= 2.8k) screen). Just in case the issue is specific to 200% display scale.

### Expected behavior

One line of text per item.

### Ubuntu release

Other (specify in the last field)

### What architecture are you using?

amd64

### System info

- AMD64
- Display is 2800 pixels wide.

PR #1021, opened 2025-04-02
fix: align remaining confirm page entries to summary table

d-loose:
image

This should now match the alignment in the Figma designs.

Fix: #1011 
UDENG-6573


matthew-hagemann (approved):
Thanks!

PR #1031, opened 2025-04-07
fix: Install libgles2 (LP: #2106101)

tobhe:
See https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2106101
It looks like on the Pinebook Pro we are hitting:
```
ubuntu@ubuntu:~$ ubuntu-desktop-bootstrap
Couldn't open libGLESv2.so.2: libGLESv2.so.2: cannot open shared object file: No such file or directory
Aborted (core dumped)
```

This seems like the obvious fix.

An additional inclusion in `prime` is not needed because `usr/lib/*/libGL*.so.*` matches `libGLESv2.so.2` and `libGLESv2.so.2.1.0`. I can add one for clarity if desired.


d-loose (approved):
Thank you, LGTM :+1:

tobhe:
Fixed mail address and signed

d-loose:
Sorry, it looks like your commit is still not [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). I can sign it for you in case you don't want to fiddle with the git(hub) config right now :)

tobhe:
Since this is a bug fix I think it would technically make sense to cherry-pick it to 24.04 too. What do you think?

d-loose:
Yes, good point - will do

PR #1037, opened 2025-04-09
chore: update subiquity for 25.04

d-loose:


matthew-hagemann (approved):
Go sanity is failing all over and doesn't impact the changes being made here, so approving

PR #1050, opened 2025-04-15
chore(provd): downgrade go deps to match main branch

d-loose:
Let's keep the golang dependencies up to date with `main`. This should make the CI checks pass again on `ubuntu/25.04`.


matthew-hagemann (approved):
Wow, a passing Go-Sanity-Check in the CI?!?

PR #1071, opened 2025-05-28
feat: move recovery key page to the end

d-loose:
This moves the existing recovery key page out of the storage wizard and to the very end of the flow. I had to split off the 'done' page from the existing `InstallPage` and move it into a new `DonePage`. The old behavior of the `InstallPage` was to simply exchange the slideshow for the 'done' page, now it advances to the next page instead.

UDENG-7046


matthew-hagemann (approved):
LGTM :slightly_smiling_face:

PR #1076, opened 2025-06-04
feat: add file picker to recovery key page

d-loose:
Adds a file picker to the recovery page to allow the user to save the key to a file.
If the user selects a directory anywhere in `/target` or a directory under a mountpoint that has the `/cow` file system as a source (i.e. the ephemeral overlay fs of the live iso), an error is raised.
The error handling UX is still WIP

UDENG-7047


d-loose:
Error handling UX:

[Screencast From 2025-06-06 14-15-32.webm](https://github.com/user-attachments/assets/68cd6e3d-9397-49e7-a987-e8f4d760ba29)

matthew-hagemann (approved):
one question, but otherwise LGTM :slightly_smiling_face:

juanruitina:
UI seems OK. I wonder if we should default the file picker to /media/[username]? I understand external drives are generally there?

juanruitina:
That seems OK; we make this more convenient when supported, not the end of the world if it isn't.

PR #1085, opened 2025-06-12
WIP: feat: expose pin/passphrase entropy for tpm/fde

d-loose:
[Screencast From 2025-06-12 14-16-55.webm](https://github.com/user-attachments/assets/4f89c68e-989c-43cc-bf91-72c1af982baf)

Inconsistent text size in the input decoration will be fixed by https://github.com/ubuntu/yaru.dart/pull/1005.

UDENG-7050


apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_widgets.dart
```diff
@@ -56,10 +55,12 @@ class _PassphraseFormFieldState extends ConsumerState<PassphraseFormField> {
...
               errorText: hasTpmLowEntropy
-                  ? model.entropy?.localize(lang, model.passphraseType)
+                  ? model.entropy?.semanticEntropy
+                      .localize(lang, model.passphraseType)
                   : null,
               helperText: hasTpmSufficientEntropy
-                  ? model.entropy?.localize(lang, model.passphraseType)
+                  ? model.entropy?.semanticEntropy
+                      .localize(lang, model.passphraseType)
                   : null,
```
matthew-hagemann, line 64: The helper text is bigger than the error hint text by default. I ran into the same issue in the [security-center](https://github.com/canonical/desktop-security-center/blob/main/packages/security_center/lib/widgets/passphrase_widgets.dart#L148) and solved it with adding: `helperStyle: Theme.of(context).textTheme.bodySmall` in the `InputDecoration`

matthew-hagemann (changes requested):
Looks great! One small nitpick

apps/ubuntu_bootstrap/lib/pages/storage/passphrase/passphrase_widgets.dart
matthew-hagemann: Ah perfect! In that case, LGTM :slightly_smiling_face:

PR #1130, opened 2025-07-18
fix(a11y): Add window control semantic labels

ashuntu:
Adds semantic labels to all window controls (the close/maximize/minimize buttons).

This issue was just reported for the installer, but it's basically the same fix for all the apps/packages, so they're included, too :).

---

Audit 1857656
UDEAA-119


d-loose:
Sorry, just re-discovered this PR in my notifications - could you please rebase it?

PR #1143, opened 2025-08-07
chore: update go deps

d-loose:
The CI complains again. We should clarify the future of `provd` at some point, since there's currently not much value in keeping its dependencies up to date besides appeasing the CI.


matthew-hagemann (approved):
Thanks for this! Yea, we should see if its popping into a future roadmap anytime soon and if not archive it.

PR #1145, opened 2025-08-07
feat(provd): Ubuntu Insights integration

hk21702:
This PR implements `provd` integration with Ubuntu Insights.

Since the philosophy of Ubuntu Insights differs from that of Ubuntu Report, an entirely new v2 telemetry service was created to prevent the introduction of any breaking changes. This exists in parallel with the old telemetry service which still utilizes Ubuntu Report.

As part of this effort, this PR includes changes to fix issues and lintian errors when building for Questing. Note that the existing change log seems to differ a bit from what is on Launchpad. This wasn't corrected in this PR in case something specific was needed.

The diff is primarily protobuf generation and copyright file changes.


hk21702:
I've just been informed that `provd` and `ubuntu-desktop-init` are no longer in use. I'll keep this PR up for now so that the work doesn't go to waste in case this part of the code base is ever revisted.

ogres:
Hi @hk21702 , I will close this PR, but let's keep the branch so that we can look into it in future when we need it. Meanwhile, since we are not going to merge it for now, better to keep the PR closed. Thanks!

PR #1163, opened 2025-08-25
fix(a11y): Add focus ring to show/hide password button

ashuntu:
Adds a focus ring around the show/hide button when focused.

image

---

UDENG-7689


ashuntu:
Note for myself: looks like ValidatedFormField doesn't have readable semantics, could use some more investigation

juanruitina:
The focus ring looks good. I couldn't help noticing though that the field showing an error message ("Your computer's name") has a dark, barely noticeable border. Should be same red as the text. Could you fix?

PR #1164, opened 2025-08-26
fix(a11y): Adjust ListTile color scheme for better contrast

ashuntu:
Adjusts the `ListTile` widget to have better contrast in different states.

Example below. From top to bottom the states shown are hovered, selected, and focused.

| | Dark | Light |
|-|------|-------|
| Default | image | image |
| HC | image | image |

---

UDENG-7683

Related to https://github.com/ubuntu/yaru.dart/issues/979

### DIFF
--- a/packages/ubuntu_provision/lib/src/widgets/themed_list_tile.dart
+++ b/packages/ubuntu_provision/lib/src/widgets/themed_list_tile.dart
@@ -3,7 +3,6 @@ import 'package:yaru/yaru.dart';
 
 class ThemedListTile extends StatefulWidget {
   const ThemedListTile({
-    required this.valueKey,
     required this.selected,
     required this.title,
     super.key,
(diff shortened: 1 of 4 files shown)


juanruitina:
This is looking great, love to see the checkmark there.

Please revert to the previous hover. Hover [doesn't require 3:1 contrast with background](https://www.tpgi.com/a-whole-lot-of-bovver-over-hover/): we shouldn't sacrifice text legibility, the user can infer the state from the mouse cursor position.

Also, the Ubuntu logo should be the negative one in dark mode. Shall we open a separate ticket for that?

packages/ubuntu_provision/lib/src/widgets/themed_list_tile.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:flutter/material.dart';
+import 'package:yaru/yaru.dart';
+
+class ThemedListTile extends StatefulWidget {
+  const ThemedListTile({
+    required this.valueKey,
```
d-loose, line 6: We don't need a separate `valueKey` here, and can just use the `key` instead. The `ListTile` further down the tree doesn't need a `key` if we have one here

d-loose (approved):
Thanks!

PR #1166, opened 2025-08-28
fix: Images not adjusting to dark mode

ashuntu:
Images (namely, the initial Ubuntu logo on the first page) in the installer do not adjust to the selected theme. This fixes the image for the Ubuntu logo, but the Landscape logo still needs a dark variant.


d-loose (commented):
Thanks, this looks great now! Could you update the `README` to include the new `image-dark` property?

PR #1180, opened 2025-09-09
ci: Avoid installing locales and man pages

adombeck:
Speed up CI jobs by avoiding installation of locales and man pages.

The major performance gain that can be expected is that it makes the "Processing triggers for man-db" step a no-op. That step can otherwise take more than a minute, for example in [this recent run](https://github.com/canonical/ubuntu-desktop-provision/actions/runs/17575083622/job/49918442057#step:5:672) it took 1m 40s:

image

UDENG-7865


.github/workflows/ci.yml
```diff
@@ -118,6 +120,7 @@ jobs:
           submodules: recursive
       - uses: asdf-vm/actions/install@v3
       - uses: bluefireteam/melos-action@v3
+      - uses: canonical/desktop-engineering/gh-actions/common/dpkg-install-speedup@main
       - run: sudo apt update
       - run: sudo apt install -y clang cmake curl gsettings-desktop-schemas libgtk-3-dev ninja-build pkg-config unzip
       - run: xvfb-run -a -s '-screen 0 1024x768x24 +extension GLX' flutter test integration_test/${{matrix.target}}
```
adombeck, line 126: I don't know if we use locales of the installed packages in the e2e tests, in which case we might not want to use the dpkg-install-speedup here.

matthew-hagemann:
@adombeck Awesome, thanks for this! Same commit signing issue I'm afraid :slightly_smiling_face:

matthew-hagemann (approved):

PR #1195, opened 2025-09-26
Add workaround for gtk #2599

snalvc:
Workaround for [gtk#2599](https://gitlab.gnome.org/GNOME/gtk/-/issues/2599)

Fix [LP: 2124253](https://bugs.launchpad.net/ubuntu-desktop-provision/+bug/2124253)

- Detect if 3.24.9 <= gtk3 version < 3.24.42 && running on wayland. If so, correct pixel size by scale
- Maximize window rather than entering fullscreen when default window size > workarea

Tested on a 10" tablet with integer scale 200% and Dell XPS 13 9340 with integer scale 200%

remmina_10 102 97 176_10 102 97 176_20250926-054617
remmina_10 102 180 95_10 102 180 95_20250926-054415

### DIFF
--- a/apps/factory_reset_tools/linux/my_application.cc
+++ b/apps/factory_reset_tools/linux/my_application.cc
@@ -1,5 +1,9 @@
 #include "my_application.h"
 
+#ifdef GDK_WINDOWING_WAYLAND
+#include <gdk/gdkwayland.h>
+#endif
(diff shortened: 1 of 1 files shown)


d-loose (approved):
Thank you, LGTM :+1:

PR #1208, opened 2025-10-08
chore: track subiquity branch ubuntu/questing

dbungert:


d-loose:
Thanks, I'll have a look at this next week - it needs some more branch shuffling and a new snap package for the 26.04 installer on launchpad. I'm tracking this in UDENG-8264.

PR #1209, opened 2025-10-10
Change the Orca pronunciation with the language

sergio-costas:
This patch is a proof-of-concept to change the pronuciation of the Orca Screen Reader when the language changes in the installer.

This only works on 25.10 onwards, thanks to the changes in Gnome Shell to fully rely on systemd.

### DIFF
--- a/packages/ubuntu_provision/lib/src/locale/locale_model.dart
+++ b/packages/ubuntu_provision/lib/src/locale/locale_model.dart
@@ -1,6 +1,8 @@
+import 'dart:io';
 import 'dart:ui';
 
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:path/path.dart' as p;
 import 'package:safe_change_notifier/safe_change_notifier.dart';
 import 'package:ubuntu_localizations/ubuntu_localizations.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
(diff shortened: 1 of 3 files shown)


sergio-costas:
https://github.com/user-attachments/assets/759a930e-7fbb-4f66-8664-9c0ce01c097c

This is how it does it in my system... I boot from an ISO image, closed the installer, installed the rebuilt snap, and relaunched the installer.

d-loose:
Doesn't matter which language I try. Maybe it's a race-condition? After manually turning the screen reader off and then on again through the a11y menu, the voice changes

d-loose:
I've build this from 93aac220a1706a42fe630b1485762b1e9363d644. The terminal in the lower left corner of my screen recording shows the output of `watch cat .config/systemd/user/orca.service.d/override.conf`, where you can see that the contents of the file are updated correctly

d-loose:
I've tested this again after increasing the duration to 1 second
```diff
diff --git a/packages/ubuntu_provision/lib/src/locale/locale_model.dart b/packages/ubuntu_provision/lib/src/locale/locale_model.dart
index ccb40cf0d..b7707a761 100644
--- a/packages/ubuntu_provision/lib/src/locale/locale_model.dart
+++ b/packages/ubuntu_provision/lib/src/locale/locale_model.dart
@@ -65,7 +65,7 @@ class LocaleModel extends SafeChangeNotifier {
         if (screenReaderEnabled) {
           await accessibilityService.setScreenReader(false);
           _log.info('Waiting for screen reader to fully disable');
-          await Future.delayed(const Duration(milliseconds: 100));
+          await Future.delayed(const Duration(milliseconds: 1000));
           _log.info('Screen reader disabled');
           await accessibilityService.setScreenReader(true);
         }
```
and this seems to work.

I've also confirmed this behavior manually in the terminal. If I modify the `override.conf` and then run:
```
gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled false;gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled true
```
the screen reader restarts, but uses the same voice.
If I run
```
gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled false;sleep 1;gsettings set org.gnome.desktop.a11y.applications screen-reader-enabled true
```
instead, it restarts and uses the new voice.
Maybe it's less better to explicitly reload the systemd  config and restart the service via
```
systemctl --user daemon-reload
systemctl --user restart orca
```
instead of relying on gsettings.

PR #1213, opened 2025-11-05
fix: Restrict protobuf to ^3.1.0

ashuntu:
grpc was pulling in protobuf 5, which ends up generating incompatible dart proto files.

Ideally we should just bump our deps, but we can't reliably do that until we can bump Flutter/dart.


ashuntu:
You can see analyze failure here: https://github.com/canonical/ubuntu-desktop-provision/pull/1166

d-loose (approved):

PR #1217, opened 2025-11-13
chore: update flutter to 3.38

d-loose:
Updates Flutter to 3.38.
I've also unpinned `yaru` and `mockito`, and upgraded to `freezed` 3.1. We should also consider upgrading to [riverpod 3](https://riverpod.dev/docs/3.0_migration), but I'd do that in a separate PR.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-flutter 3.38.0-stable
+flutter 3.38.1-stable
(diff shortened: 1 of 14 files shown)


ashuntu (approved):
LGTM

Tested screen reader on Wayland and X11 and it works!

PR #1229, opened 2025-11-25
fix: factory-reset-tools: bump base snap to core24

snalvc:
Bump base snap of `factory-reset-tools` to `core24`.

This upgrade enables the use of newer `mesa-2404`, providing support for newer Intel graphics on OEM platforms.

Resolves: JIRA WTN-313


d-loose (approved):
Thanks, LGTM :+1:

PR #1230, opened 2025-11-25
feat: add functional UI for TPM action page

d-loose:
Adds a new page to the installer that displays TPM/FDE related errors to the user and allows them to perform actions to fix those errors in some cases.

Depends on https://github.com/canonical/subiquity/pull/2269

[Screencast From 2025-11-25 16-40-51.webm](https://github.com/user-attachments/assets/edaec5f6-ae44-4dfa-b569-43ac3792493c)

Screencast shows the flow in dry-run mode, where the action fixes the error instantly. In reality, the user would have to perform an action that reboots the system at some point.

UDENG-8610


apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_page.dart
```diff
@@ -0,0 +1,90 @@
...
+    final model = ref.watch(tpmActionModelProvider);
+    final notifier = ref.read(tpmActionModelProvider.notifier);
+
+    final children = [
+      if (model.tpmError != null) Text(model.tpmError!.kind.localize(lang)),
+      if (model.confirmationNeeded)
+        YaruCheckButton(
+          value: model.confirmed,
+          onChanged: (value) => model.confirmed = value!,
+          title: Text('confirm'),
```
matthew-hagemann: I think this should be an l10n sting?

matthew-hagemann:
I also see that once you have expanded the advanced options in the Encryption and file system page, you can't revert to not using TPM/FDE:
[Screencast From 2025-11-28 16-07-10.webm](https://github.com/user-attachments/assets/9eed6c0f-a950-42d0-95ab-179442b9838f)

d-loose:
> I also see that once you have expanded the advanced options in the Encryption and file system page, you can't revert to not using TPM/FDE: [Screencast From 2025-11-28 16-07-10.webm](https://github.com/user-attachments/assets/9eed6c0f-a950-42d0-95ab-179442b9838f)

That's an artifact of the `tpm.yaml` test source, which isn't very realistic. It only offers the `CORE_BOOT_ENCRYPTED` capability (that also means it's implicitly selected by default even if you don't press the 'advanced options' button). In reality there would always be at least the `DIRECT` capability shown on that page as well

PR #1249, opened 2026-01-09
feat: implement UI structure from Figma for action API

d-loose:
This updated the UI structure of the TPM action page to match the latest Figma designs. There are still a lot of missing l10n strings for the error and action descriptions, which I'll add as soon as they're available.
Also, there are some UI issues (font sizes in the warning box and the checkbox tile, text overflow in the check box tile, etc), which we should address in upstream yaru.

Screenshot From 2026-01-09 10-17-00
Screenshot From 2026-01-09 10-18-03
Screenshot From 2026-01-09 10-18-09
Screenshot From 2026-01-09 10-19-13

UDENG-8684

### DIFF
--- a/apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_page.dart
+++ b/apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_page.dart
@@ -99,6 +99,7 @@ class _ActionBody extends ConsumerStatefulWidget {
 
 class _ActionBodyState extends ConsumerState<_ActionBody> {
   late bool isConfirmed;
+  SubiquityException? error;
 
   @override
   void initState() {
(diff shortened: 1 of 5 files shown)


apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_page.dart
```diff
@@ -100,16 +82,91 @@ class TpmActionPage extends ConsumerWidget with ProvisioningPage {
...
+              value: isConfirmed,
+              onChanged: (value) => setState(() => isConfirmed = value!),
+              title: Text(widget.action.warning!.confirmationLabel(lang)),
+            ),
+          ],
+          Row(
+            children: [
+              OutlinedButton(
+                onPressed: isConfirmed
+                    ? () => notifier.performAction(widget.action)
```
matthew-hagemann: Not in this code change, but I ran this without updating `subiquity`, and I see if we get an error returned in the `tpmActionModel` it is unhanded and we get stuck in a loading state.

Could we add a try-catch to `performAction` and set an error to show?

PR #1302, opened 2026-02-05
fix(a11y): Add toggle state to the install logs toggle button

ashuntu:
The toggle button for install logs now will convey its pressed state to screen readers.

---

UDENG-8925


d-loose (approved):
Thanks!

PR #1316, opened 2026-02-19
feat: update l10n strings for tpm fde action page

d-loose:
* Update existing error descriptions
* Add hints to find options in firmware settings
* Use singular l10n string if there's only one action and expand tile by default

image

UDENG-8937


matthew-hagemann (dismissed):
LGTM!

matthew-hagemann (approved):

PR #1324, opened 2026-02-27
feat: update mascot for resolute

d-loose:
This includes the new artwork and also adds support for dark variants of image assets used in the slideshow.

UDENG-9066


matthew-hagemann (approved):
LGTM!

PR #1326, opened 2026-03-01
feat: Add age verification compliance for CA AB-1043 and CO SB26-051

dylanmtaylor:
Implements per-user age verification signals required by:
- California AB-1043 (Digital Age Assurance Act)
- Colorado SB26-051 (Age Attestation on Computing Devices)

Changes:
- Add gRPC service with username-based age data storage
- Add D-Bus API accepting username parameter for age queries
- Store age data per-user in /var/lib/AccountsService/users/<username>.age
- Add Flutter UI page for age verification during provisioning
- Provide age brackets: under 13, 13-15, 16-17, 18+

Multi-user support:
- Each user has separate age data file
- Applications query by username


ArrayBolt3:
It's worth noting that the original spec concept I posted on the XDG and ubuntu-devel mailing lists is almost certainly not what will be standardized across all distros in the end. The current most-likely-to-be-used implementation to my awareness is in xdg-desktop-portal and is already being worked on by KDE and GNOME: https://github.com/flatpak/xdg-desktop-portal/pull/1922

Whatever implementation ends up being used in the end (if any, it's entirely possible no one will implement this depending on how things go), it would need to be more flexible than this when it comes to age brackets to be adaptable to whatever jurisdictions implement similar laws in the future.

PR #1333, opened 2026-03-04
feat: implement customizable horizontal layout for error page

d-loose:
This updates the error page to use the horizontal layout that is used on most other pages, and adds an error parsing method for customization.
I've also added a custom error parser for autoinstall errors:

[Screencast From 2026-03-04 13-42-57.webm](https://github.com/user-attachments/assets/ba03ecb7-a190-4358-8b96-09fa4aa6f58c)

UDENG-8914

### DIFF
--- a/apps/ubuntu_bootstrap/lib/app/installer_wizard.dart
+++ b/apps/ubuntu_bootstrap/lib/app/installer_wizard.dart
@@ -155,10 +155,6 @@ class _AutoinstallWizard extends ConsumerWidget {
         InstallationStep.error.route: WizardRoute(
           builder: (_) => ErrorPage(
             allowRestart: false,
-            errorParser: (e) => errorParser(
-              e,
-              UbuntuBootstrapLocalizations.of(context),
-            ),
           ),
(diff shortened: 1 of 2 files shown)


matthew-hagemann:
> Updated:
> Screencast.From.2026-03-05.14-58-20.webm
> 
> @matthew-hagemann do you remember if there was a reason to remove the padding from the confirmation page when displaying the autoinstall contents?
> 
> https://github.com/canonical/ubuntu-desktop-provision/blob/ce6beb72033cd2b85e5760f91f742db4137304fc/apps/ubuntu_bootstrap/lib/pages/confirm/confirm_page.dart#L81-L82
> 
> Changed in #982

@d-loose I see in the PR I wrote that the margins ended up being too large with the autoinstall confirmation text, but I don't remember the specifics :thinking: 

Looking at the screen shot from then though, the margins are different to whats in your screen recording, so not sure if something else has changed padding wise elsewhere? Maybe my workaround can be safely reverted?

Current:
image

My PR:
image

PR #1338, opened 2026-03-06
identity: add birthDate field to user provisioning [DRAFT]

dylanmtaylor:
## Motivation

Age verification legislation (California AB-1043, effective Jan 1 2027; Brazil SB26-051 / Lei 15.211/2025) requires OS providers to collect date of birth during account/device setup and expose an API for age brackets.

## Changes

**Proto**: add `birth_date` string field to User message (field 6).

**Backend (provd)**: validate birthDate (YYYY-MM-DD, year ≥ 1900, not future), call `SetBirthDate` on the AccountsService user object via D-Bus after user creation.

**Frontend (Flutter)**: add date picker widget to the identity page, wire birthDate through Identity model and gRPC bridge. The form cannot be submitted without a valid birth date.

**Bootstrap (live installer)**: persist `BirthDate` to `/tmp/bootstrap-postinst.conf` via `PostInstallService` so the postinst hook (#1339) can write it to the target.

**Localization**: machine translations added for all 75 locales. These should be reviewed by native speakers before release.

## Dependencies

- #1339 — postinst hook to write BirthDate to target AccountsService keyfile (targets `snap/ubuntu-desktop-bootstrap/main`)
- [accountsservice MR](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/merge_requests/176) — `BirthDate` property + `SetBirthDate` D-Bus method

## Integration

- [systemd PR #40954](https://github.com/systemd/systemd/pull/40954) — `birthDate` in JSON user records
- [xdg-desktop-portal PR #1922](https://github.com/flatpak/xdg-desktop-portal/pull/1922) — portal API for age brackets

## Testing

- End-to-end tested on Ubuntu 25.10 VM: gRPC CreateUser → provd validates and calls SetBirthDate via D-Bus
- provd unit tests pass
- Flutter app builds successfully

image
image


AdrianVovk:
You can't just drop a user record into /run/userdb. The same user record will be provided by two userdb services: NSS and drop-in. Thus, they'll race. If drop-in wins, then it gets to return the user's complete user record (which means all the settings stored in NSS will be gone!). If NSS wins, then the settings stored in the drop-in will be discarded (i.e. the birthdate)

For now, we use accountsservice on the desktop. That means someone will need to add an API to accountsservice similar to the systemd API. AccountsService should just transparently pass things through to userdb when interacting with a homed user. And in the non-homed case, it'll store the setting directly in /var/lib/accountsservice. When we transition off of accountsservice, we'll need to migrate all the settings in /var/lib/accountsservice into userdb anyways, and birthDate will be one of them.

dylanmtaylor:
> an2dlh
> 

You really need a hobby or something

PR #1375, opened 2026-03-26
a11y(ubuntu_bootstrap): announce installation progress and completion

### LINKED ISSUE #1373: A11Y: Orca should report install status/progress
Currently when you have gone through all the install steps and the installation is actually taking place there is no indication to Orca what is happening and when it is finished.
We should find a way to make Orca aware of whats happening.

paul-geoghegan:
Currently, screen readers like Orca do not automatically read the installation progress steps or the final completion status when the screens load. 

Due to known limitations with Flutter's Linux AT-SPI2 accessibility bridge (specifically around declarative liveRegion updates and header traits), passive semantics are often dropped. 

This PR addresses this by explicitly firing imperative announcements to the screen reader:
- Converted InstallPage and DonePage to ConsumerStatefulWidgets.
- InstallPage: Uses ref.listen to imperatively announce InstallationAction changes and a post-frame callback for the initial state.
- DonePage: Uses a post-frame callback to announce the final Installation Complete status immediately upon page load without hijacking the user's keyboard focus.


d-loose (commented):
Thanks, this looks great!
Would you mind removing the pubspec changes, though - I'll bump the lower version bounds in a separate PR

paul-geoghegan:
@d-loose there you go. Sorry, I was being silly.

d-loose (approved):
Great, thanks a lot!

PR #1394, opened 2026-04-08
fix: update telemetry file when selecting a `guidedCapability`

d-loose:
We recently added a new telemetry option to detect TPM/FDE installations in #1384. While testing this I noticed that we never see any telemetry regarding 'advanced features' such as TPM/FDE, ZFS, etc. It looks like this is due to #917 where the old 'advanced features' dialog was moved to a dedicated page. The call to the telemetry service is still made on the 'storage' page, so later changes to the `guidedCapability` aren’t reflected in the telemetry file.

This PR adds an additional call to the telemetry service to ensure the `PartitionMethod` is updated when the `guidedCapability` is changed. It also adds a dedicated option for encrypted ZFS and fixes the OEM status (which was always set to `false`).
I've added corresponding checks to the integration test as well.

UDENG-9638


matthew-hagemann (approved):
LGTM!

PR #1426, opened 2026-04-30
feat: add support for desktop-agnostic keyboard layout persistence

eeickmeyer:
As ubuntu-desktop-bootstrap was built to be used by all Ubuntu flavors, we need to make sure we're not using too many GNOME-specific idioms. Otherwise desktops, such as KDE Plasma, miss out on certain features like setting the keyboard layout. Currently, this is completely missed when one uses this installer on Ubuntu Studio.

This pull request makes the keyboard layout setting more DE-agnostic.

### DIFF
--- a/apps/ubuntu_bootstrap/lib/services/keyboard_service.dart
+++ b/apps/ubuntu_bootstrap/lib/services/keyboard_service.dart
@@ -68,13 +68,19 @@ class SubiquityKeyboardService implements KeyboardService {
     @visibleForTesting GSettings? inputSourceSettings,
     @visibleForTesting
     Future<ProcessResult> Function(String, List<String>)? runProcess,
+    @visibleForTesting
+    Future<void> Function(String layout, String variant)? setLocale1X11Keyboard,
   })  : _inputSourceSettings = inputSourceSettings ??
             createService<GSettings, String>('org.gnome.desktop.input-sources'),
-        _runProcess = runProcess ?? Process.run;
+        _runProcess = runProcess ?? Process.run,
+        _setLocale1X11KeyboardImpl =
+            setLocale1X11Keyboard ?? _defaultSetLocale1X11Keyboard;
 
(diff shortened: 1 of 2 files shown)


apps/ubuntu_bootstrap/lib/services/keyboard_service.dart
```diff
@@ -110,29 +117,74 @@ class SubiquityKeyboardService implements KeyboardService {
...
+      name: 'org.freedesktop.locale1',
+      path: DBusObjectPath('/org/freedesktop/locale1'),
+    );
+    final args = [
+      DBusString(layout), // x11Layout
+      const DBusString(''), // x11Model
+      DBusString(variant), // x11Variant
+      const DBusString(''), // x11Options
+      const DBusBoolean(false), // convert (do not propagate to vconsole)
+      const DBusBoolean(true), // interactive (allow polkit agent prompt)
```
ogres, line 179: I am unsure about this...
Installer app runs without privileges.
If we pass here false, it will fail due to missing privileges.
If we pass here true, it will ask for user/password - which would be a weird user experience since user/password has not yet been set up.

The right place to have this is probably within subiquity (the underlying system that Installer communicates with) at  https://github.com/canonical/subiquity/blob/9f3558984a2aafc08398d3ada8cadfba1ad29de7/subiquity/server/controllers/keyboard.py#L249

Since subiquity runs with privileges, it will be able to directly set it without additional prompts. 

@eeickmeyer Could you please check this on subiquity ?

eeickmeyer: Alright, sounds like a plan. I opened https://github.com/canonical/subiquity/pull/2427. Perhaps we should close this one?

ogres: Yes, that would make sense. Thank you for contribution!

PR #1438, opened 2026-05-14
feat: announce current wizard step to screen readers

d-loose:
The wizard's step indicator (the dots at the bottom of each page) is
currently invisible to screen readers. This PR wraps it in a `Semantics`
widget so assistive technology announces the current position, e.g.
*"Step 2 of 5"*.

## What changed

- **`ubuntu_wizard`** — added a `stepSemanticsLabel` parameter and a
  static `defaultStepSemanticsLabel` builder to `WizardBar`. When a
  label is provided, the page indicator dots are wrapped in
  `Focus` + `Semantics` with `ExcludeSemantics` on the children (so
  individual dots aren't read separately). Steps are 0-indexed
  internally but announced as 1-based.

- **`ubuntu_bootstrap`** — added a localized `stepIndicatorLabel`
  string (`"Step {CURRENT_STEP} of {TOTAL_STEPS}"`) and wired up
  `WizardBar.defaultStepSemanticsLabel` once in `_InstallerApp.build`,
  so all pages get the behavior without individual call-site changes.

### DIFF
--- a/packages/ubuntu_wizard/lib/src/wizard_bar.dart
+++ b/packages/ubuntu_wizard/lib/src/wizard_bar.dart
@@ -7,6 +7,41 @@ import 'package:yaru/yaru.dart';
 /// The spacing between Continue and Back buttons.
 const kWizardBarSpacing = 8.0;
 
+/// Scopes a default step-semantics-label builder to all descendant
+/// [WizardBar] widgets that do not provide their own
+/// [WizardBar.stepSemanticsLabel].
+///
+/// Place this above the navigator (or any common ancestor of all [WizardBar]
+/// widgets) to apply a localized label without modifying each call site:
+/// ```dart
+/// WizardBarTheme(
+///   stepSemanticsLabel: (step, total) =>
+///       l10n.stepIndicatorLabel(step, total),
+///   child: ...,
+/// )
+/// ```
+class WizardBarTheme extends InheritedWidget {
+  const WizardBarTheme({
+    super.key,
+    required this.stepSemanticsLabel,
+    required super.child,
+  });
+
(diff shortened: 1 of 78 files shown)


d-loose:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/ubuntu-desktop-provision/pull/1438#pullrequestreview-4290305806)

packages/ubuntu_wizard/lib/src/wizard_bar.dart
paul-geoghegan, line 1: When you press the next button there is anouncement which I like but it's wrong. For example it will say step 1 of 12 instead of 2 of 12.

paul-geoghegan (dismissed):
Things seem way better but need a small change.

packages/ubuntu_wizard/lib/src/wizard_bar.dart
d-loose: Interesting, that actually seems to be related to an existing bug: when pressing the 'next' button and loading the next page takes more than 100ms, the button is disabled and a loading spinner is shown. This causes the button to become unfocusable, so the previous widget is focused. Without the changes from this PR, the reader announces 'back button' as soon as the 'next' button loses its focus when entering a loading state. In the context of this PR that means that the semantic label from the page indicator is announced once the 'next' button enters a loading state.
What's the best solution here? i.e. what behavior do you expect for the 'next' button in this case? Should it keep the focus? Should the loading state be announced?

paul-geoghegan: That is probably out of scope for this PR then. LGTM in that case and we can discuss further on the related issue.

paul-geoghegan (approved):
LGTM

PR #1439, opened 2026-05-15
feat: Add accessibility labels to Wi-Fi list tiles for screen readers

d-loose:
Currently, the Wi-Fi network list conveys signal strength and encryption
status only through icons (wifi signal icon with/without a lock), which
is inaccessible to screen reader users. The connection state
(connected/connecting) is similarly only shown via icons.

This PR adds Semantics widgets around the relevant icons in `WifiListTile`
so that screen readers can announce:

- Signal strength — one of: No signal, Weak signal, OK signal, Good
  signal, Excellent signal
- Security — Open network or Secure network
- Connection state — Connected or Connecting…

### Changes

- Added 9 new l10n keys: `networkWifiSignalNone` through
  `networkWifiSignalExcellent`, `networkWifiOpenNetwork`,
  `networkWifiSecureNetwork`, `networkWifiConnecting`,
  `networkWifiConnected`
- Wrapped the trailing wifi icon in
  `Semantics(label: _wifiSemanticsLabel(...))`
- Wrapped the leading status icon (checkmark/spinner) in `Semantics`
  with the appropriate connection state label
- Added `_wifiSemanticsLabel()` helper that maps
  `AccessPoint.strength` and `AccessPoint.isOpen` to localized strings

### DIFF
--- a/packages/ubuntu_provision/lib/src/network/wifi_view.dart
+++ b/packages/ubuntu_provision/lib/src/network/wifi_view.dart
@@ -255,12 +255,13 @@ class WifiListTile extends ConsumerWidget {
   }
 }
 
+int _wifiStrength(AccessPoint model) => (model.strength ~/ 20 + 1).clamp(1, 5);
+
 String _wifiSemanticsLabel(
   AccessPoint model,
   UbuntuProvisionLocalizations lang,
 ) {
-  final strength = (model.strength ~/ 20 + 1).clamp(1, 5);
-  final strengthLabel = switch (strength) {
+  final strengthLabel = switch (_wifiStrength(model)) {
     1 => lang.networkWifiSignalNone,
     2 => lang.networkWifiSignalWeak,
     3 => lang.networkWifiSignalOk,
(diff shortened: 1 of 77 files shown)


d-loose:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/ubuntu-desktop-provision/pull/1439#pullrequestreview-4296884976)

paul-geoghegan:
@d-loose I'm getting no additional output rom Orca when trying to review this. The icons probably aren't being read since I'm not atually focusing on them. You may have to use merge semantics on the container or something similer to get it to read correctly.

packages/ubuntu_provision/lib/src/network/wifi_view.dart
paul-geoghegan, line 1: Okay, this now works perfectly, The only thing we could maybe improve is not duplicating the logic for determining the network strength but it's fine as is if you want it like that.

paul-geoghegan (commented):
All good besides the one comment.

packages/ubuntu_provision/lib/src/network/wifi_view.dart
d-loose: good point, I've refactored it a bit

matthew-hagemann (approved):
LGTM

PR #1450, opened 2026-06-02
feat: add missing autoinstall error cases

d-loose:
This is a follow-up to https://github.com/canonical/ubuntu-desktop-provision/pull/1333, where we've updated the error page design and added a specific message for error cases where a user-supplied autoinstall command fails (`AutoinstallUserSuppliedCmdError`). This PR adds a error messages for the more missing `AutoinstallError`, `AutoinstallValidationError` and `CloudInitSchemaValidationError` cases, e.g.:
- `AutoinstallError`:
  ```yaml
  version: 1
  timezone: Invalid/Timezone
  ```
  Screenshot From 2026-06-02 16-37-24
- `AutoinstallValidationError`
  ```yaml
  version: 1
  identity: foobar
  # missing 'password' and 'hostname' fields
  ``` 
  Screenshot From 2026-06-02 16-36-28
- `CloudInitSchemaValidationError`
  ```yaml
  version: 1
  user-data:
    ssh_import_id: lp:test
    # ssh_import_id expects an array, so it should be '[ lp:test ]' here
  ``` 
  Screenshot From 2026-06-02 16-47-26

See also https://github.com/canonical/subiquity/blob/main/subiquity/ui/views/error.py#L454-L471

UDENG-9922


matthew-hagemann (dismissed):
LGTM!

matthew-hagemann:
I see the go: sanity check is failing again, I'll open a PR to fix it so it doesn't create noise in future PR's

apps/ubuntu_bootstrap/lib/l10n/ubuntu_bootstrap_en.arb
```diff
@@ -44,6 +44,9 @@
   "autoinstallLandscapeOption": "Automated with Landscape",
   "autoinstallLandscapeDescription": "For users in organizations that provide an autoinstall file via Landscape.",
   "autoinstallErrorMessage": "A command in the autoinstall file failed to run during installation.",
+  "autoinstallGenericErrorMessage": "The installation will be unable to proceed with the provided autoinstall file.",
```
juanruitina: Use "Installation cannot continue with the provided autoinstall file."

juanruitina:
OK, I went with Ana through this. In addition to the string change I suggested, can I ask for two more changes?

- [ ] Move the primary button right above the collapsible. That way we interrupt the flow and prevent the user from clicking where the "Next" button usually is impulsively.
- [ ] Please revert the formatting of the error message under technical details. Given it's a proper string, such formatting is not needed, sorry for the back and forth. Can we make sure it's selectable?

It should look like this:

image

d-loose:
> It should look like this:

This means we should also make the following l10n string changes, right?
- "Installation failed" -> "Something went wrong"
- "Restart installer" -> "Restart installation"

juanruitina:
@d-loose Keep the strings as they are, I updated the mockup to revert that (sorry, was based on an older design).

d-loose:
image

juanruitina:
Fantastic, UX+1! Thank you for the changes :)

matthew-hagemann (approved):

PR #1476, opened 2026-06-25
feat: added Yaru focus boarders to all interactive controls

### LINKED ISSUE #1475: A11Y: Focus rings should be used for all interactive elements
Currently we have focusrings disabled in our Yaru theme so that they could be consistant with the default Flutter theme. We should revert this and wrap any custom widgets in the Yaru Focus Boarder widget in order to have better focus state behavior and improve a11y for sighted keyboard users.

paul-geoghegan:
This PR adds support for the Yaru focus boarders. Currently we are relying on the native Flutter material focus backgrounds which provide very poor contrast and fail WCAG guidelines. This PR enables them for all Yaru widgets and wrapps all non Yaru interactive widgets with Yaru's focus boarder widget.


d-loose:
Thanks for working on this! There are now also focus borders for non-interactive elements such as page headers and step indicators - is this intended? It also looks like there's now an extra focus border for the `OptionButton`, which already has one built-in, so I suppose that's redundant.

@juanruitina since this is quite a drastic visual change, could you take a look at the screen recording below and provide some feedback?

[Screencast From 2026-07-15 13-01-10.webm](https://github.com/user-attachments/assets/1f36a57f-bd4f-4da5-a6f0-becac062a594)

paul-geoghegan:
@d-loose yes I talked to @juanruitina and we agreed that the text should have a focus border since it's focusable.
In terms of the redundant one I can remove that if you wish. I guess I miss-understood that one in particular.

juanruitina:
Bit of context: the problem is that non-interactive elements are focusable at all, but I understand that's necessary for making them readable by screen readers on Flutter. Any focusable element should show a focus ring when focused, or else experience will quickly become confusing for users that use keyboard navigation but not a screen reader.

That said, @paul-geoghegan can you double check the OptionButton focus ring? It _should_ have the ring, but in @d-loose's recording shows some weird behaviour:

- The focus ring is there, but without the 2px offset/padding we discussed (and you showed me)
- I notice the default border of the component (in the default and selected states) seems to change to 2px thickness and adds border radii. Both borders seem to be clashing somehow

Another minor comment: I'm not sure the focus ring for "back" and "next" buttons have the 2px offset.

Happy to debug this visually over a call (once I'm back).

paul-geoghegan:
Fixed both issues — turned out `YaruFocusBorder`'s default `strokeAlign` already gives the 2px offset on its own, so I'd accidentally doubled it by also passing `borderPadding`. Removed that. Also found the inner radio button wasn't passing `hasFocusBorder: false`, so it was still drawing its own redundant ring even though it can't actually receive focus — that's fixed too.

Back/Next buttons had the same doubled-offset issue, fixed the same way.

Verified visually on the storage/install-type pages, offset and borders look right now on both selected and unselected tiles.

d-loose:
Thanks, it looks like there's still a duplicate focus ring around the `OptionButton`s, though:

[Screencast From 2026-08-20 12-41-57.webm](https://github.com/user-attachments/assets/5a1ae6c8-01ee-4b2b-8fa0-57f707365a58)

paul-geoghegan:
Thanks for catching that. The duplicate ring was caused by `YaruRadioListTile` already drawing its own focus ring internally (via its `YaruListTile`), and we had additionally wrapped the whole tile in a `YaruFocusBorder` — so two rings appeared on focus. Removed the extra outer wrap; the ring is now the one built into `YaruRadioListTile`.

One known limitation: the built-in ring in yaru 10.0.0 uses `strokeAlignInside` so it sits flush against the tile edge with no gap. This was fixed upstream in yaru 10.2.0 (the internal tile now tracks focus state and uses an outside-aligned ring with proper spacing). When yaru gets bumped in this repo the offset will appear automatically with no further changes needed on our side.

d-loose:
> When yaru gets bumped in this repo the offset will appear automatically with no further changes needed on our side.

I've just bumped yaru to 10.2.0 on `main` - do you mind rebasing your branch?

d-loose:
Now the focus rings are missing entirely for the `OptionButton`s

[Screencast From 2026-08-21 14-50-31.webm](https://github.com/user-attachments/assets/40fb2eb3-654b-4ee0-8ffa-581bc7064d2a)

d-loose (approved):
Thanks, LGTM now!

PR #1485, opened 2026-07-01
feat: enforce PIN/passphrase encryption for TPM/FDE when HWROT is missing

d-loose:
In order to support TPM/FDE on systems that don't have a hardware root of trust, users will be required to setup a PIN or passphrase during the installation. This PR includes the necessary changes in the UI:
- obtain `CoreBootEncryptionRequirement`s from subiquity
- remove 'Unlock disk automatically' as needed
- adapt l10n strings as needed

[Screencast From 2026-07-01 17-08-11.webm](https://github.com/user-attachments/assets/121022c4-9bc9-4cbb-a8e1-0fe986a37885)

@juanruitina the 'Additional Security' page is still the same as the one shipped in 25.10. Looking at your new designs for 26.04.1, I see a few differences:
- "Require a PIN/passphrase" -> "Require a PIN/passphrase on startup". Should I use the new string here?
- The subtitles have been removed. Should I remove them entirely, or only in the case where PIN/passphrase is enforced?
- The subtitles for PIN/passphrase read "More secure. .." and "Most secure. ..", respectively, in the previous designs. However, we currently show the options in the order provided by subiquity/snapd (see screencast). Should we enforce an order (PIN before passphrase) in the UI?

Depends on:
- https://github.com/canonical/subiquity/pull/2391
- https://github.com/canonical/subiquity/pull/2394
- https://github.com/canonical/snapd/pull/17289

UDENG-10586


juanruitina:
Designs are now accurate; a bit has changed since we first implemented this, so there are some minor changes here and there. Subiquity sorting is fine.

- [x] Please check all strings match. Yes, let's drop the subtitles; they add too much complexity. I also updated the paragraph in the "Set an encryption passphrase/PIN".
- [x] Link in "Additional security" should point to https://ubuntu.com/desktop/docs/en/latest/explanation/hardware-backed-disk-encryption/#tpm-fde-encryption-passphrase
- [x] Please move the "Learn more" links below the options, as per the designs.

d-loose:
With HWROT:
Screenshot From 2026-07-02 17-14-54

No HWROT:
Screenshot From 2026-07-02 17-17-09
image

juanruitina:
LGTM, UX+1! I still think the additional security page shouldn't be visible when there is HWRoT for a more streamlined installation, but that's a separate discussion.

matthew-hagemann (approved):
LGTM!

PR #1493, opened 2026-07-14
feat: disable pin/pass for TPM/FDE unless explicitly required

d-loose:
As requested by the design team.


matthew-hagemann (approved):

juanruitina:
For the record: before this PR, the installer would show the passphrase and PIN page if Subiquity communicated they were supported; in practical terms, always, regardless of whether passphrase or PIN were _required_ (currently when there is no hardware root of trust).

UX-wise, this was a regression: TPM alone is good enough security-wise for a vast majority of the users when there is root of trust, and using PIN and passphrase defeat the purpose of the feature in terms of convenience. We should also keep the installer simple; users will always be able to set up passphrase and PIN in the Security Center.

PR #1499, opened 2026-07-21
fix(tpm-action): only consider errors for the selected disk

d-loose:
When attempting to do a TPM/FDE install while another disk which isn't large enough to accommodate the OS is present, the TPM action page currently prevents the user from proceeding. This happens because disallowed targets for all disks (instead of just the selected disk) are checked.

Example response from `/storage/v2/guided` in that case:
```json
{
  "status": "DONE",
  "error_report": null,
  "configured": null,
  "targets": [
    {
      "disk_id": "disk-nvme0n1",
      "ptable": null,
      "allowed": [
        "DIRECT",
        "LVM",
        "LVM_LUKS",
        "ZFS",
        "ZFS_LUKS_KEYSTORE",
        "CORE_BOOT_PREFER_ENCRYPTED"
      ],
      "disallowed": [],
      "$type": "GuidedStorageTargetReformat"
    },
   ...
    {
      "disk_id": "disk-mmcblk0",
      "ptable": null,
      "allowed": [],
      "disallowed": [
       ...
        {
          "capability": "CORE_BOOT_PREFER_ENCRYPTED",
          "reason": "TOO_SMALL",
          "message": null,
          "errors": null
        }
      ],
      "$type": "GuidedStorageTargetReformat"
    },
    ...
  ]
}
```

This PR addresses this by explicitly checking only the disallowed capabilities of the selected target.
I also noticed that we don't update the selected target if it changes after performing a `PROCEED` action (the lists for `allowed` and `disallowed` capabilities will be updated). It seems like subiquity doesn't look at those when posting to `/storage/v2/guided`, but I think we should send the correct data nonetheless.

UDENG-11056

### DIFF
--- a/apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_model.dart
+++ b/apps/ubuntu_bootstrap/lib/pages/storage/tpm_action/tpm_action_model.dart
@@ -58,10 +58,17 @@ class TpmActionModel extends SafeChangeNotifier {
 
   Future<void> _updateGuidedStorage() async {
     final response = await _storage.getGuidedStorage();
+    final selectedTarget = _storage.guidedTarget;
+    if (selectedTarget is! GuidedStorageTargetReformat) {
+      _disallowedTarget = null;
+      return;
+    }
     _disallowedTarget = response.targets
         .whereType<GuidedStorageTargetReformat>()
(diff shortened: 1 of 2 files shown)


gabelluardo (approved):
LGTM, thanks!

PR #1501, opened 2026-07-21
feat: update l10n strings for PIN/passphrase

d-loose:
Update the l10n strings for the PIN/passphrase page as discussed.


juanruitina (approved):
UX+1!

PR #1503, opened 2026-07-22
test: add integration test case for TPM/FDE w/o HWROT

d-loose:
Adds an integration test using the test data from https://github.com/canonical/subiquity/pull/2404 to assert that setting up a PIN or passphrase is enforced when there's no HWROT.


gabelluardo (approved):
thanks!

PR #1526, opened 2026-08-18
fix(network): guard wifiDeviceProvider against empty device list

marktantongco:
## Problem

`wifiDeviceProvider` indexes the Wi-Fi device list with `devices[index]` and no bounds check:

```dart
final wifiDeviceProvider = Provider.family<WifiDevice, int>(
  (ref, index) => ref.watch(wifiModelProvider).devices[index],
);
```

`WifiView` guards the empty case for the widget tree (`if (model.devices.isEmpty) return const SizedBox.shrink();`), but that does **not** protect the Riverpod provider itself. When `wifiModelProvider` notifies a change, Riverpod re-evaluates the already-active `wifiDeviceProvider(0)` instance *before* the watching widget is disposed. If the list is momentarily empty, `devices[0]` throws:

```
RangeError: RangeError (length): Valid value range is empty: 0
#0  wifiDeviceProvider.<anonymous closure>
```

This crashes the Ubuntu 26.04 desktop installer (`ubuntu-desktop-bootstrap` snap, rev 589) on the "Connect to the network" screen when the Wi-Fi adapter briefly transitions to `unavailable` (e.g. during a scan), which empties the device list.

## Fix

Return `null` when the index is out of range and skip the tile in that case.

## Testing

- `flutter analyze` clean for the package.
- Manually: with the adapter flapping to `unavailable`, the provider no longer throws and `WifiView` renders the empty state instead of crashing.

Fixes a crash observed in the field (see https://github.com/marktantongco/ubuntu-26.04-installer-bootfix for the full diagnosis).

### DIFF
--- a/packages/ubuntu_provision/lib/src/network/wifi_view.dart
+++ b/packages/ubuntu_provision/lib/src/network/wifi_view.dart
@@ -110,8 +110,11 @@ class _WifiViewState extends ConsumerState<WifiView> {
   }
 }
 
-final wifiDeviceProvider = Provider.family<WifiDevice, int>(
-  (ref, index) => ref.watch(wifiModelProvider).devices[index],
+final wifiDeviceProvider = Provider.family<WifiDevice?, int>(
+  (ref, index) {
+    final devices = ref.watch(wifiModelProvider).devices;
+    return index < devices.length ? devices[index] : null;
+  },
 );
 
 class WifiListView extends ConsumerWidget {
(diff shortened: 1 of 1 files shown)


ogres:
@marktantongco Thanks for the contribution. 
We should probably not use `index` here at all, and instead have an unique identifier for each device and use that for watching, for example `device.hwAddress`. 
Current (old) logic assumes that the order of the devices do not change, which might not be true in some edge cases.

Could you please adjust that? That way, we will be naturally be forced to handle null-cases as well.

PR #1528, opened 2026-08-20
chore: resolute: bump flutter to 3.44.0 on riscv64

vhaudiquet:
Also drop tool-patch as 3.44.0 has tool upstream support for riscv64


d-loose (approved):
Thanks!

PR #1531, opened 2026-08-29
refactor: Use built-in providers for installer model

spydon:
Replaces the `InstallerModel` `ChangeNotifier` with two generated riverpod providers, `applicationStatusProvider` and `hasRouteProvider`, so the installer status stream and route lookup no longer need a manually initialized and disposed model.

- `installer_model.dart`: `InstallerModel` removed in favour of `@riverpod` providers (`restartProvider` unchanged).
- `installer_wizard.dart` and `confirm_page.dart`: read the status through `applicationStatusProvider` and routes through `hasRouteProvider`.
- `test_confirm.dart`: `buildConfirmPage` overrides `applicationStatusProvider` directly instead of stubbing the service.
- `installer_model_test.dart`: tests the providers through a mocked `InstallerService`, including subscription cleanup on dispose.

Part of #717
Supersedes #718, which was rebuilt from scratch on current `main` since the old branch predates the move to `apps/ubuntu_bootstrap`.

### DIFF
--- a/apps/ubuntu_bootstrap/lib/app/installer_model.dart
+++ b/apps/ubuntu_bootstrap/lib/app/installer_model.dart
@@ -1,42 +1,16 @@
-import 'dart:async';
-
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:safe_change_notifier/safe_change_notifier.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_bootstrap/services.dart';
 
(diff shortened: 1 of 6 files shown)


ogres (dismissed):

ogres:
@spydon Changes look good, but we probably need to update the branch to avoid conflicts with https://github.com/canonical/ubuntu-desktop-provision/pull/1532

spydon:
> @spydon Changes look good, but we probably need to update the branch to avoid conflicts with #1532

I merged in latest main (it could've been cleanly merged into main already though).

ogres (approved):

PR #1532, opened 2026-08-29
fix: Pin html below 0.15.7 and regenerate mocks

spydon:
CI is currently red for every new PR, for two reasons that are both caused by `packages/*` having no lockfile:

- `html 0.15.7` (published 2026-08-28) removed the `matches()` helper from `query_selector.dart` that `flutter_html 3.0.0` still calls, so nothing in `ubuntu_provision` or `ubuntu_wizard` compiles under `flutter test` anymore. This pins `html` below `0.15.7` in those two packages, which are the only `packages/*` members that pull in `flutter_html`. The apps are unaffected because their lockfiles pin `html 0.15.6`.
- The mock generator now emits a blank line after the `ignore_for_file` header, so `flutter-mocks` and `subiquity-types` fail their outdated-files check on 27 files. This commits the output of `melos generate`.

No functional changes. Split out of #1531 so it can land on its own.

https://claude.ai/code/session_01K4ipmNnmKt7PB21TSeVtHK

### DIFF
--- a/packages/landscape_client/test/landscape_client_test.mocks.dart
+++ b/packages/landscape_client/test/landscape_client_test.mocks.dart
@@ -3,6 +3,7 @@
 // Do not manually edit this file.
 
 // ignore_for_file: no_leading_underscores_for_library_prefixes
+
 import 'dart:async' as _i7;
 
 import 'package:grpc/service_api.dart' as _i2;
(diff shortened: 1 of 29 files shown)


ogres (approved):
Thank you 👍

PR #1533, opened 2026-08-31
fix(factory_reset): handle cli reboot failure by ptyxis with a message

medicalwei:
Running `factory-reset-tools.cli` in Ptyxis (Default terminal emulator in 26.04) to reboot into Factory Reset can fail, due to Ptyxis inhibiting logout when a program is running inside terminal.

This proposed "fix" is to provide an appropriate message that asks user to reboot the system manually.

To turn off the behavior on Ptyxis:
```
gsettings set org.gnome.Ptyxis inhibit-logout false
```

Internal Reference: OEX86-972


ogres:
@medicalwei We should probably use the localized message instead of hardcoded one for this.
Let's also see the possibility for allowing the actual restart.

ogres:
@medicalwei I see that change was merged in snapd, do we still need this PR?

medicalwei:
I'll open another for the actual fix instead.

PR #1540, opened 2026-09-08
feat: update mascot for 26.10

d-loose:
Update the mascot svg assets for 26.10. The files provided by the brand team don't have any margins built-in, so I've added an `imagePadding` property to the `HorizontalPage` which is set to match the size of the right margin for the try-or-install page. For screenshots and further discussion see jira.

BRND-3365


gabelluardo (dismissed):
Thanks for the update! 📈 

Ready to merge on a green CI.

gabelluardo (approved):

PR #1541, opened 2026-09-08
fix: use `NativePlatform` instead of deprecated `LocalPlatform`

d-loose:
`flutter analyze` fails in CI due to the recent [`platform`](https://pub.dev/packages/platform) update (its version is not pinned in `ubuntu_provision`). This PR migrates to the new classes.

### DIFF
--- a/apps/factory_reset_tools/pubspec.lock
+++ b/apps/factory_reset_tools/pubspec.lock
@@ -401,10 +401,10 @@ packages:
     dependency: transitive
     description:
       name: platform
-      sha256: "5d6b1b0036a5f331ebc77c850ebc8506cbc1e9416c27e59b439f917a902a4984"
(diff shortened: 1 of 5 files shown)


ogres (approved):
Should we also pin it to avoid similar issues in future?

d-loose:
I'm wondering what would be the proper way to do that? Currently the lock files of the application packages in `apps/` are version tracked, while those in `packages/` are not (see https://dart.dev/tools/pub/private-files#pubspec-lock)

ogres:
> I'm wondering what would be the proper way to do that? Currently the lock files of the application packages in `apps/` are version tracked, while those in `packages/` are not (see https://dart.dev/tools/pub/private-files#pubspec-lock)

Maybe we do not commit whole .lock file, but in dependencies list directly mention the version which we tested / worked in the past. In the end `app` will create the overall .lockfile

PR #1544, opened 2026-09-09
chore: update go

d-loose:

PR #1546, opened 2026-09-09
Updated golangci-lint

ogres:
Trying out new version of `golangci-lint` since the pinned one is no longer compatible in connection with https://github.com/canonical/ubuntu-desktop-provision/pull/1544


d-loose (approved):
Thanks!

